PR #65, opened 2023-02-28
Justin krs/customer detail base page setup

justinengle-krs:
The pull request includes most of the Customer Detail page.

-  Get customer info on load and display as per zeplin designs

Zeplin: 
image

Implementation:
image

- Allow tab navigation to multiple routes on the page

Zeplin:
image

Implementation::
image

- Allow adding of address through address modal

Zeplin:
image

Implemenation:
image

- List addresses and display as per zeplin
Zeplin:
image

implementation
image

PR #66, opened 2023-02-28
Justin krs/customer nav

justinengle-krs:
The PR includes the Navigation functionality which exists on the Customer Page.

- Create a nav component to render out the tabs
- Create all components for each tab route
- Create story


src/components/Customer/Customer.svelte
```diff
@@ -4,25 +4,45 @@
...
+    { name: 'Segments', component: Segments },
+    { name: 'Data Sync', component: DataSync },
+    { name: 'Dynamic Pricing', component: DynamicPricing }
+  ];
+
   async function loadData() {
     loading = true;
     try {
       customer = await getCustomerById(customerId);
+      selectedComponent = tabs[0].component;
```
matthampton-stock2shop, line 37: Is the selected tab not something that we'd want to include in the url route? (feels to me like that would be useful)... Or is it there already?

src/components/Customer/Tabs.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
```
matthampton-stock2shop, line 1: Could we maybe add a story for the Tabs component?

matthampton-stock2shop (commented):
I'm going to merge this in the #65 and then link to these comments from there

src/components/Customer/Tabs.svelte
justinengle-krs: There is one?

CustomerTabs.stories.svelte?

src/components/Customer/Customer.svelte
justinengle-krs: Let me see what can be done, it's currently just tabs on the` customer/{id}` and defaults to details.

But it should be possible for something like `customer/{id}/orders` or `customer/{id}?tab=Orders`

src/components/Customer/Tabs.svelte
matthampton-stock2shop: Just assuming we plan on re-using it - so having a place with 'generic' parameters might be nice - but I guess not necessary - don't worry

PR #67, opened 2023-02-28
Justin krs/customer detail

justinengle-krs:
Render customer details control on customer page

- Do some CSS organizing
- Add getUserById to display username
- Create a reusuable TextInput component
- Use class models on deserializing to allow for better control
- Create story


src/models/Customer/customer.ts
```diff
@@ -1,8 +1,9 @@
-import type { Meta } from '@storybook/addon-svelte-csf';
+import type { Meta } from '../meta';
 import type { Source } from '../source';
 import type { User } from '../user';
+import { Address } from './address';
 
-export interface SystemCustomer {
+export class SystemCustomer {
```
matthampton-stock2shop, line 6: Is there a reason you decided to change these from interfaces to classes?

matthampton-stock2shop (commented):
Going to merge back in to #66 and then on in to #65.

src/models/Customer/customer.ts
justinengle-krs: Mostly a preference and standard we use.
I much prefer having more explicit control over how it deserializes from the JSON in the constructor and handling nulls. Do you have a preference?

matthampton-stock2shop: No - that's fine by me

PR #69, opened 2023-02-28
Justin krs/address list

justinengle-krs:
Render out addresses, that are listed on a customer.

- Create a control
- Bind to client.addresses
- Style according to Zeplin
- Fix some incorrect properties in other scaffolding tab routes


src/components/Customer/Details/Addresses.svelte
```diff
@@ -15,6 +16,7 @@
 
   function addAddress(): void {
     customer.addresses.push(newAddress);
+    customer.addresses = customer.addresses;
```
matthampton-stock2shop, line 19: These seems a bit odd - is this to kick off a re-render? Wouldn't it be better to create a new array (I'm assuming these props are meant to be more or less immutable)?

matthampton-stock2shop (commented):
This is going to get merged down all the way in to #65 before getting merged in to master

PR #74, opened 2023-03-05
Justin krs/active dropdown

justinengle-krs:
Add client active/inactive dropdown to top of page and bind to client

Zeplin design:
image

Implementation:
image

image


matthampton-stock2shop:
@justinengle-krs Does zeplin have an example of the Inactive one? (Just wondering about the capitalization - 'InActive' looks  a bit odd to me)

flumonion:
> @justinengle-krs Does zeplin have an example of the Inactive one? (Just wondering about the capitalization - 'InActive' looks a bit odd to me)

`Inactive` looks right to me, without camel case.

src/components/Customer/CustomerDetail.svelte
```diff
@@ -67,7 +68,7 @@
   $: isLoading = $customerData.isLoading || $userData.isLoading || $customerUsersData.isLoading;
   $: userData.setOptions(['user-get', userId], () => getUserById(userId), { enabled: !!userId });
 
-  $: isReady = customer != null && user != null && customerUser != null && !isLoading;
+  $: isReady = customer != null && user != null && customerUser != null && !isLoading && customerUserVersionControlled?.workingCopy;
```
matthampton-stock2shop, line 71: Is it possible to not have a 'workingCopy'? (once it is loaded)

src/components/Customer/IsActiveDropdown.svelte
```diff
@@ -0,0 +1,27 @@
+<script lang="ts">
+  export let isActive: number;
+  export let id = `id_${Math.floor(Math.random() * 100 * Date.now())}`;
+
+  let options = [
+    { id: 1, value: 'Active' },
+    { id: 0, value: 'InActive' },
```
matthampton-stock2shop, line 7: As I said - the camel-case looks a bit funny to me - was this in the design spec? If not, I'd prefer "Inactive"

justinengle-krs: @matthampton-stock2shop  Not that I could see, wasn't sure what it should say. I fixed the casing :)

src/components/Customer/CustomerDetail.svelte
justinengle-krs: It's the version we'd want to bind to though? so we know if the user made changes

src/stories/components/Customers/IsActiveDropdown.stories.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+  import { Meta, Story } from '@storybook/addon-svelte-csf';
+  import IsActiveDropdown from '@/components/Customer/IsActiveDropdown.svelte';
+</script>
+
+<Meta title="components/Customers/IsActiveDropdown" component={IsActiveDropdown} />
```
justinengle-krs, line 6: @matthampton-stock2shop I have some troubles opening a default select via JS - using the play you showed me. Have you managed to do this?

src/components/Customer/CustomerDetail.svelte
matthampton-stock2shop: I'm happy to merge this in to #73 - and then we can carry on the conversation there rather

src/stories/components/Customers/IsActiveDropdown.stories.svelte
matthampton-stock2shop: Right... Let me give it a bash

src/components/Customer/CustomerDetail.svelte
```diff
@@ -105,6 +106,7 @@
 {#if isReady}
   <div>
     <PageHeader title="{customer.first_name} {customer.last_name}" hasBackButton={true}>
+      <IsActiveDropdown bind:isActive={customerUserVersionControlled.workingCopy.active} />
```
matthampton-stock2shop, line 109: Is there a reason you're not using `customerUser.active` here?

justinengle-krs: The  customerUserVersionControlled object references itself i.e workingCopy, updatedVersion and original to know the state of the data. So workingcopy is the version that would bind to the 'form'. Do you have a different idea?

matthampton-stock2shop: I just ask because you have some easier to read variables already set up that I think look nicer in the actual component code:

```
$: customerUser = (customerUserVersionControlled && customerUserVersionControlled.workingCopy) || null;
```

matthampton-stock2shop: i.e. I think it will look a bit ugly if `customerUserVersionControlled.workingCopy.blah` is scattered all over the file - when there is a nice alias for it already

justinengle-krs: Ah yes, so I suspect once I do the "Save changes" portion, that these will have to go. I don't think the 2 way binding will work through the svelte reactive...

So like $customerUser -> form is fine
but it needs to go form->$customerUserVersionControlled.workingCopy.prop

src/stories/components/Customers/IsActiveDropdown.stories.svelte
matthampton-stock2shop: This does indeed look tricky to do - I think the native select component doesn't let itself be remote controlled in a visual way (you can fire events to change the current selection - but not to 'open' the component) - so happy to just leave that out.

PR #83, opened 2023-03-14
Handle Unauthorized request

justinengle-krs:
If API response is 401. Navigate away and invalidate token.

To test:

1. Login as per normal
2. In a new tab login again into the v1 app https://console.stock2shop.com/, this should invalidate the API TOKEN
3. Click back to the v2 app. Perform any action that makes an api call, you should get logged out. This should happen automatically on any svelte-query screen i.e customer-detail


src/lib/api/api.ts
```diff
@@ -1,9 +1,15 @@
+import { push } from 'svelte-spa-router';
 import { apiConfigInstance, authRequestInterceptor } from '@/lib/api/apiConfigInstance';
+import { s2sStore } from '../s2sStore';
 import type { APIRequest, RequestOptions } from './types';
 
 // the response is a bit unpredictable, tricky to manage types here
 async function handleResponse<T>(response: Response): Promise<T> {
   try {
+    if (response.status === 401) {
+      await push('/');
```
matthampton-stock2shop, line 10: Does it make sense to do this before setting the token to null? (probably no biggie - just interested)?

justinengle-krs: Setting the token to null automatically causes App.svelte to display the Login page, however the pages seems to end up in a broken state once authed again (Svelte doesn't reload it properly)

So to resolve this, I'm navigating away first and then invalidating the token.

matthampton-stock2shop: Ok - thought it might be something like that!

matthampton-stock2shop (approved):

PR #85, opened 2023-03-16
Justin krs/orders

justinengle-krs:
Orders List tab
https://trello.com/c/DYmvh9ko/38-orders-list
https://trello.com/c/7rvzdFzb/41-orders-search
https://trello.com/c/jwvsjGsH/42-order-multiselect

**Notes:**
- As discussed with Matt, the styling on the order status is not viable or worth it, as this is a random string based on the channel.
- Column configurator not needed according to Chris.

**Chromatic build**
https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=26

**Zeplin:**
image

**Implementation:**
image


src/env/local/persistent.ts
```diff
@@ -89,6 +51,13 @@ export const persistentConfig: ConfigPersistent = {
         'customer.last_name',
       ],
     },
+    [ElasticSearchIndex.CUSTOMER_ORDERS]: {
```
justinengle-krs, line 54: We don't want the column configurator  on orders screen to affect the customer/orders screen

src/components/Customer/Orders.svelte
```diff
@@ -1,10 +1,159 @@
...
-  Render orders for : {customer.id} <br />
-  {customer.first_name} - {customer.last_name}
-</div>
+{#if $data.isError}
+  <h1>Failed to load customer data.</h1>
+{:else if isLoading}
+  <Loader />
+{:else}
+  <div>
+    <PageHeader title="{customerVersionControlled.workingCopy.first_name} {customerVersionControlled.workingCopy.last_name}" hasBackButton={true}>
```
TheSuperRealJP, line 119: Is the intended plan to repeat this PageHeader for each one of the Customer tabs? It feels like this belongs on Customer > Details.svelte. It might be out of scope of this PR, but worth considering before this header is repeated again for Attributes, Segments, Data Sync, etc.

src/components/Customer/Orders.svelte
```diff
@@ -1,10 +1,159 @@
...
+  async function saveChanges() {
+    if (customerVersionControlled.isStale && !confirm('The customer information has changed, by saving you will override these changes.')) {
+      return;
+    }
+
+    loading = true;
+
+    const apiResult = await customersRepo.update(new SystemCustomerDataWrapper({ system_customer: customerVersionControlled.workingCopy }));
+    if (!apiResult.result) {
+      alert(apiResult.error);
```
TheSuperRealJP, line 83: I know this may be moved based on my later comment, but does this "alert" does not stop the flow of `saveChanges` when `!apiResult.result`. You'll still get a `alert('Customer saved');` which is wrong.

TheSuperRealJP (changes requested):
Orders.svelte inherits too much repeated code (HTML &  TS) from CustomerDetail.svelte.

justinengle-krs:
> Orders.svelte inherits too much repeated code (HTML & TS) from CustomerDetail.svelte.

Good point. I have made changes to resolve this.

CustomerMaster.svelte has been created. It contains all logic and requirements for handling the container of the page:
image

CustomerDetail and Orders only does whats specific to that portion.

CustomerMaster implements its own SaveChanges which Saves the customer, and then fires off a passed in delegate function to execute any custom save logic for the current page

src/components/Customer/Orders.svelte
justinengle-krs: covered in reduce duplications

justinengle-krs:
> @justinengle-krs FYI
> 
> Issue: saving customer while on tab other than "Details" causes blank screen afterwards.
> 
> Reproduce steps:
> 
> 1. Go to /#/customer/5676648/details
> 2. Save changes and confirm page reloads just fine after save.
> 3. Go to Orders tab.
> 4. Save changes and confirm spinner animation and success alert.
> 5. See blank page. Browser "back" also doesn't do anything (URL is updated though)
> 
> Cause: Likely this error in console image

Good catch thanks. Resolved

TheSuperRealJP (approved):

src/components/Customer/CustomerDetail.svelte
```diff
@@ -17,12 +17,10 @@
   import type ChannelUsersRepo from '@/repos/channel-users-repo';
```
matthampton-stock2shop, line 17: Chromatic change expected? https://www.chromatic.com/test?appId=63ef6cddab1dec384359e1cd&id=641c33b2d963e7d76245eddc

src/routes/CustomerOrders.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  import Orders from '@/components/Customer/Orders.svelte';
+  import CustomersRepo from '@/repos/customers-repo';
+
+  export let params: { id: number };
+
+  const customersRepo = new CustomersRepo();
+</script>
+
+<Orders customerId={params.id} {customersRepo} />
```
matthampton-stock2shop, line 10: Could we add a storybook screen for this?

e.g. to show something like the one you have in the description

image

src/components/Customer/CustomerDetail.svelte
justinengle-krs: In the zeplin the warning colour seems to be #f5931c. So this is correct?

See here: https://app.zeplin.io/project/629fb41be1a2021104a7cae0/screen/63f8ebaf47d0ec1035743b7b

src/env/common/columns/orderColumns.ts
```diff
@@ -17,7 +22,7 @@ const orderColumns: TableColumn[] = [
   },
   {
     field: 'channel_order_code',
-    label: 'Channel Order Code',
```
matthampton-stock2shop, line 20: I think we should leave this label as 'Channel Order Code' (is it is not the 'Channel' - it is a code for each each order e.g. #2123)..

(And we are happy for this to be the column that is showing rather than Channel as in the design)

src/components/Customer/CustomerDetail.svelte
matthampton-stock2shop: Oops- I was seeing the green 'diff' highlighting and it looked like the color had changed to green!! Sorry!

src/env/local/persistent.ts
matthampton-stock2shop: Makes sense - I'm assuming somewhere there must be some knowledge that the CUSTOMER_ORDERS is not an actual elastic search index though (i.e. that it needs to query the ORDERS index on the backend) .. Can you point me to that code? (Sorry if it was obvious)

matthampton-stock2shop (changes requested):

justinengle-krs:
@matthampton-stock2shop Can't seem to reply on the ElasticSearch question. On Customer/Orders.svelte we use ElasticSearchIndex.ORDERS to query the API

matthampton-stock2shop:
> @matthampton-stock2shop Can't seem to reply on the ElasticSearch question. On Customer/Orders.svelte we use ElasticSearchIndex.ORDERS to query the API

Ok - so when is ElasticSearchIndex.CUSTOMER_ORDERS actually used then?

justinengle-krs:
> > @matthampton-stock2shop Can't seem to reply on the ElasticSearch question. On Customer/Orders.svelte we use ElasticSearchIndex.ORDERS to query the API
> 
> Ok - so when is ElasticSearchIndex.CUSTOMER_ORDERS actually used then?

s2sStore.ts is where its used to retrieve the columns. 

There's a mapColumns which seems to "inner join" the system config with the persistent config.

Both which have a "key" for CUSTOMER_ORDERS

See: /common/local/system.ts and /common/local/persistent.ts

src/env/local/persistent.ts
matthampton-stock2shop: Oh, ok - got it

src/routes/CustomerOrders.svelte
justinengle-krs: DOne

src/env/common/columns/orderColumns.ts
justinengle-krs: Done

matthampton-stock2shop (approved):

PR #89, opened 2023-03-29
Justin krs/order detail-Customer details and notes only [1]

justinengle-krs:
Order detail:
**customer details and notes only**
Chromatic: https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=31
Zeplin:
image

Implemenation:
image


src/stories/routes/85-Order.stories.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
```
matthampton-stock2shop, line 1: I made a couple of comments on the chromatic build:

https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=31

matthampton-stock2shop: (see the red ones)

src/models/sources/source.ts
```diff
@@ -17,8 +17,22 @@ export default class SystemSource {
 
   source_id?: number;
 
+  source_customer_code: string;
+
+  source_order_code: string;
```
matthampton-stock2shop, line 22: This seems a bit odd to me - do we put these values in to a 'source' object in an order? Maybe we can rather make a subclass of this called "order_source" or something like that? And use that one for an order...

matthampton-stock2shop: In fact! In our database this is a different table, called - "order_source"

image

It doesn't have all of the other properties of a source, though - it is just a link - and stores the idenfifier of this order for that source.

So I think it makes more sense to create a new model for this - rather than extending the existing source model....

matthampton-stock2shop (changes requested):

src/models/customer/customer.ts
```diff
@@ -68,6 +70,8 @@ export default class SystemCustomer {
   }: Partial<SystemCustomer> = {}) {
     this.id = id;
     this.created = created;
+    const createdDate = new Date(created);
+    this.createdDisplay = `${createdDate.toDateString()} ${createdDate.toLocaleTimeString()}`;
```
matthampton-stock2shop, line 74: Also - does it make sense for this to be in the model? It feels like a view responsibility to me...

src/models/sources/source.ts
justinengle-krs: Makes sense, will sort out

matthampton-stock2shop: @justinengle-krs Could we remove these properties from the source model, since they are now in the new one?

src/stories/routes/85-Order.stories.svelte
matthampton-stock2shop: Hi @justinengle-krs .. I think this second comment on chromatic is still relevant:

"Could we add an example where the customer create date is not undefined"

https://www.chromatic.com/test?appId=63ef6cddab1dec384359e1cd&id=642425e4af7fb60f02bb683b

src/models/sources/source.ts
justinengle-krs: Sorted

src/stories/routes/85-Order.stories.svelte
justinengle-krs: https://www.chromatic.com/test?appId=63ef6cddab1dec384359e1cd&id=642bfaa0c87eb930b45a0549

matthampton-stock2shop (approved):

PR #93, opened 2023-04-05
Casper krs/dynamic pricing

casperlotter:
Create dynamic pricing contracts for a specific customer.

https://app.zeplin.io/project/629fb41be1a2021104a7cae0/screen/63bbff67d4712453b8eb2939
https://trello.com/c/LOGEtsXy/69-dynamic-contract-pricing-post
https://www.stock2shop.com/help/guides/contract-pricing/


src/models/meta.ts
```diff
@@ -5,7 +5,7 @@ export interface Meta {
   template_name?: any;
   created: string;
   modified: string;
-  client_id: string;
+  customer_id: string;
```
justinengle-krs, line 8: is it always customer_id? Maybe another endpoint uses this mordel and requires both?

casperlotter: @justinengle-krs Meta is unused outside of the DynamicPricing code at this point. types s2s/types import it but never reference it. So, it is safe to make this change.

src/models/pricing-contracts/pricing-contract.ts
```diff
@@ -0,0 +1,111 @@
+import type { Meta } from '../meta';
+
+function createMeta({ id, customerId, key, value }: { id: number | null; customerId: string; key: string; value: string }): Meta {
```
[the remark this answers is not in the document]
casperlotter: Meta is a data type and does not currently have custom behaviour. Let us leave it like this for now,

src/models/pricing-contracts/pricing-contract.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.type = type;
+    this.attributeValue = attributeValue;
+    this.value = value;
+    this.id = id;
+    this.customerId = customerId;
+    this.comparison = comparison;
+  }
+}
+
+export function loadContracts(meta: Meta[]): PricingContract[] {
```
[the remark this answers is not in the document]
casperlotter: I am fine with leaving it here. This code will be used together and having a separate file will not be useful. It can possibly move to the customer repo but I feel it is more out of place there than here.

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,237 @@
...
+
+{#if isLoading}
+  <div>Loading customer data...</div>
+{:else if $customerData.isError}
+  <div>Error loading customer data.</div>
+{:else}
+  <CustomerMaster selectedTab="dynamic-pricing" {customerVersionControlled} {customersRepo}>
+    <DynamicPricing bind:open {addContractModel} {customer} {customerMetaRepo} {updateListener} {inlineNotification} {deleteContract} />
+    {#if $inlineNotification}
+      <InlineNotification type={$inlineNotification.type}>{$inlineNotification.message}</InlineNotification>
```
justinengle-krs, line 139: The padding needs to be applied here as well

justinengle-krs (dismissed):

matthampton-stock2shop:
@casperlotter Having had a look at #124 , I wanted to check if you have done anything about the 'Order' column (either setting / editing it or re-ordering by drag-n-drop) - cos that case will be even more unpleasant - cos it might change the keys of a bunch of entries ...

If you have not yet done anything about that, I think it would be good to chat before spending a lot of time on it - so let's get this complete and then re-visit.

And then, for now, I'd like to suggest that we change the Edit to only allow the amount to be changed
(which will not change the key).

matthampton-stock2shop (changes requested):
See https://github.com/stock2shop/console_v2/pull/93#issuecomment-1540220197

casperlotter:
@matthampton-stock2shop The ordering was not touched in this. We will create a separate task for that since this is already too big a PR. I have changed the update to now use PUT as you requested and now only allow the value to change, and it works fine.

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,233 @@
...
+    open = true;
+  };
+  let inlineNotification: Writable<EventNotification | null> = writable(null);
+  let open: boolean = false;
+
+  const customersRepo = new CustomersRepo();
+  const customerMetaRepo = new CustomersMetaRepo();
+  let customerData: any;
+  let getData = () => {
+    customerData = useQuery(['customer-get', params.id], () => customersRepo.get(params.id), {
```
matthampton-stock2shop, line 44: I'd like to see a more standard use of the svelte-query here - I'm not sure of the implications of wrapping this in another callable and would prefer to use it as designed...

(i.e. you always call useQuery - and then respond to the isLoading/isError flags etc)

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,233 @@
...
+        return isDescending ? -result : result;
+      }),
+    ];
+  };
+
+  pageNumber.set(1);
+  $: isLoading = $customerData.isLoading;
+  $: {
+    if ($updateListener) {
+      $updateListener = false;
```
matthampton-stock2shop, line 119: If you change to using svelte-query in the usual way, then rather than using this 'updateListener' - which feels a bit flaky / race-conditiony to me - you should be able to use `queryClient.invalidateQueries` to get it to refetch after saving changes

src/routes/Customers.svelte
```diff
@@ -88,31 +82,26 @@
...
       selectedColumnIds={$configPersistent.elasticSearch.customers.columnIds}
       onSelected={s2sStore.setCustomerColumns}
+      onReset={() => {}}
     />
   </div>
   <div class="h-full grow overflow-y-scroll rounded">
     <div class="flex p-4">
+      <!-- search_fields={$configPersistent.customers_search_fields} -->
       <Search
-        search_fields={$configPersistent.customers_search_fields}
```
matthampton-stock2shop, line 100: Why this change?

src/models/meta.ts
casperlotter: @matthampton-stock2shop That makes sense. Now that I am working on the product pricing contracts it has crystalized for me. Will update.

matthampton-stock2shop:
https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=72

src/components/Customer/DynamicPricing.svelte
```diff
@@ -1,10 +1,59 @@
...
+        customerId: customer.id,
+      })
+    );
+  }
+
+  async function addContract(contract: PricingContract) {
+    // Ensure there is no existing contract with the key. We do not have an id reference on metas
+    let verb = contract.newContract ? 'added' : 'updated';
+    $updateListener = null;
+    const result = await customerMetaRepo.put(contract.toMetaObject());
```
matthampton-stock2shop: I'm getting an error here trying to add a new contract - it looks like the backend might want you
to use POST if it is a new one, and PUT to update an existing one...

image

matthampton-stock2shop: image

src/stories/components/100-PricingContracts/AddPricingContractModal.stories.svelte
```diff
@@ -0,0 +1,46 @@
+<script lang="ts">
+  import { Meta, Story } from '@storybook/addon-svelte-csf';
+  import AddPricingContractModal from '@/components/PricingContracts/AddPricingContractModal.svelte';
+  import PricingContract from '@/models/pricing-contracts/pricing-contract';
```
matthampton-stock2shop, line 4: e.g. 

In the following screen: 1. The 'No results' text looks a little off in terms of alignment (do we need that text?), 2. the rounded drop-shadow at the bottom looks a bit funny to me there, too - I don't think that is part of the design is it?:

image

src/stories/components/100-PricingContracts/AddPricingContractModal.stories.svelte
```diff
@@ -0,0 +1,46 @@
...
+  import { QueryClient, QueryClientProvider } from '@sveltestack/svelte-query';
+  import productJson from '@/stories/test-data/products/es_products.json';
```
matthampton-stock2shop, line 6: And error messages seem to do something odd to the add button:

image

src/routes/Customers.svelte
casperlotter: @matthampton-stock2shop Another type fix:

Property 'customers_search_fields' does not exist on type 'ConfigPersistent'.ts(2339)

matthampton-stock2shop (changes requested):
Hi @casperlotter ... Sorry for the drip-feed of comments ... I'm still working my way through this!

src/stories/components/100-PricingContracts/AddPricingContractModal.stories.svelte
matthampton-stock2shop: I'm also thinking that we should probably just display "Error saving contract" in the UI

And log the full error message to the console instead - it looks a bit nasty and developer-y like that!

src/components/Customer/DynamicPricing.svelte
casperlotter: @matthampton-stock2shop I updated the code to incorporate the customer meta and fixed this bug in the process as well as the error. I introduced this bug when I added the update functionality.

src/routes/Customers.svelte
matthampton-stock2shop: Ah

src/models/meta.ts
```diff
@@ -11,3 +11,7 @@ export interface Meta {
   required?: any;
   type?: any;
 }
+
+export interface CustomerMeta extends Omit<Meta, 'id' | 'client_id'> {
```
matthampton-stock2shop, line 15: You should be able to keep the id and client_id here - as those a properties of all meta - once they've been saved on the server (they are populated there)...

So I think better to just make them optional on the Meta type

src/components/PricingContracts/AddPricingContractModal.svelte
```diff
@@ -0,0 +1,136 @@
...
+  import { writable } from 'svelte/store';
+  import type { Writable } from 'svelte/store';
+
+  export let doPostElasticSearch: (path: string, body: DSL, params: APIRequestParams) => Promise<SystemProductsResponse>;
+  export let open: boolean;
+  const index = ElasticSearchIndex.PRODUCTS;
+  const queryClient = useQueryClient();
+  const isQueryFetching = useIsFetching(index);
+
+  const { request, isDataInvalidated } = useElasticSearchContext<SystemProductsResponse>({
```
matthampton-stock2shop, line 29: This should also move down to the FacetSearch

src/components/PricingContracts/AddPricingContractModal.svelte
```diff
@@ -0,0 +1,136 @@
...
+  const facetsWithOperators = ['price', 'created'];
+  let filterValue = writable($contract.entityAttribute);
+  let showOperators = false;
+  let verb = 'Add';
+
+  $: {
+    showOperators = facetsWithOperators.some((operator) => $filterValue.toLowerCase().includes(operator));
+
+    if ($contract.serverId !== prevContract) {
+      filterValue.set($contract.entityAttribute);
```
matthampton-stock2shop, line 78: What is this about? It seems to set a value on the search component if the contract id is different to the last time? Confused ...

src/components/PricingContracts/AddPricingContractModal.svelte
```diff
@@ -0,0 +1,136 @@
...
+
+  const operators = [
+    { label: 'Is equal to', value: '==' },
+    { label: 'Is greater than', value: '>' },
+    { label: 'Is less than', value: '<' },
+    { label: 'Is greater than or equal to', value: '>=' },
+    { label: 'Is less than or equal to', value: '=<' },
+  ];
+
+  const facetsWithOperators = ['price', 'created'];
```
matthampton-stock2shop, line 69: This code (and the  showOperators etc below) doesn't seem to be used? 

If it isn't could we remove it?

src/routes/Customers.svelte
```diff
@@ -46,17 +40,17 @@
     request: { path: esRoutes.searchCustomers },
     isFetching: isQueryFetching,
     facets: customerFacets,
-    onInit: (context: IElasticContext<SystemCustomersResponse>) => {
```
matthampton-stock2shop, line 49: I think you can restore this - now that we have a type for it again

src/components/PricingContracts/FacetSearch.svelte
```diff
@@ -0,0 +1,79 @@
...
+</script>
+
+<div class="z-20 text-left">
+  <TextInput placeholder="Search product attribute" {onInput} bind:value={$filterValue} className="mb-4 w-full" />
+  {#if isLoading}
+    <div class="flex justify-center">
+      <div class="loader ease-linear rounded-full border-4 border-t-4 border-gray-200 h-6 w-6 mb-4" />
+    </div>
+  {:else if showOptions}
+    <div class="max-h-56 overflow-y-auto my-2">
```
matthampton-stock2shop, line 69: This component seems to be trying to be a filtering select component - but in a slightly unusual way (instead of the options showing above the other content, it pushes the rest of the content down)...:

image

@justinengle-krs don't we have some examples of the more familiar filtering select components somewhere that we can steal from? 

But... I think I'm going to log this as a follow-up - cos this pull request is already a bit too big, isn't it!

matthampton-stock2shop: And in actual fact, I think we need to make it a general purpose 'Attribute Selection' widget - cos we are doing something similar for the filter controls, and will need the same for the bulk edit screens

src/components/PricingContracts/AddPricingContractModal.svelte
```diff
@@ -0,0 +1,136 @@
...
+</script>
+
+<SideModal {disabled} bind:open saveText="{verb} Contract" saveClicked={doUpdateContract}>
+  <div class="flex" slot="title"><span class="ml-[16px]">{verb} Dynamic Pricing Contract</span></div>
+  <div slot="content">
+    <div class="px-8 py-8 border-b border-neutral-200">
+      <p class="text-neutral-500 text-regular font-bold mb-4">If the following product attribute:</p>
+
+      {#if $contract.newContract}
+        <FacetSearch
```
matthampton-stock2shop, line 100: This needs to be required - at the moment you can add a contract with a blank value

src/components/PricingContracts/AddPricingContractModal.svelte
```diff
@@ -0,0 +1,136 @@
...
+        options={operators}
+        display_func={(o) => o.label}
+        className={'mb-4'}
+        valueProperty="value"
+        bind:value={$contract.comparison}
+        placeholder="Please select"
+        disabled={true}
+      />
+
+      <TextInput placeholder="Enter value" isDisable={!$contract.newContract} bind:value={$contract.attributeValue} className="mb-4 w-full" />
```
matthampton-stock2shop, line 123: This should also be required

matthampton-stock2shop: image

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,213 @@
...
+  import LocalSortHeader from '@/lib/es/LocalSortHeader.svelte';
+  import type { Order } from '@/lib/es/types';
+
+  export let params: { id: number };
+  const queryClient = useQueryClient();
+  let customerVersionControlled: ObjectVersionControl<SystemCustomer>;
+  let customer: SystemCustomer;
+  let contracts: PricingContract[] = [];
+  let pageContracts: PricingContract[] = [];
+  let perPage = 20;
```
matthampton-stock2shop, line 27: Can we make this 10? I think 20 is looks a bit too many for me:

image

(I have a very big screen - but have to scroll down to get to the pagination controls)

matthampton-stock2shop, line 27: Also - could you add a storybook screen that has more than one page - so we can see it there as well?

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,213 @@
...
+  let inlineNotification: Writable<EventNotification | null> = writable(null);
+  let open: boolean = false;
+
+  const customersRepo = new CustomersRepo();
+  const customerMetaRepo = new CustomersMetaRepo();
+  let customerData: any;
+  customerData = useQuery(['customer-get', params.id], () => customersRepo.get(params.id), {
+    onSuccess(data) {
+      customer = new SystemCustomer(data);
+      contracts = loadContracts(customer.meta, params.id);
```
matthampton-stock2shop, line 47: If you have a custom sortField then you should re-sort here (cos, for example, if you navigate away and then come back if will refresh the data) - (currently what happens is that there is a custom sort header still showing, but the data is not actually sorted)

src/routes/DynamicPricing.svelte
```diff
@@ -0,0 +1,213 @@
...
+            </div>
+          {/each}
+        {/if}
+      </div>
+      <LocalPagination {perPage} items={contracts} prev={pageFunc} next={pageFunc} />
+    </div>
+  </CustomerMaster>
+{/if}
+
+<style>
```
matthampton-stock2shop, line 198: Do we have notes to switch this out for tailwind?

matthampton-stock2shop (changes requested):
I think I've looked at everything now ... sorry for the amount of feedback!

src/routes/DynamicPricing.svelte
matthampton-stock2shop: Done by casper - needs a check

src/routes/DynamicPricing.svelte
matthampton-stock2shop: Done

src/stories/components/100-PricingContracts/AddPricingContractModal.stories.svelte
matthampton-stock2shop: Thanks for taking out the No results.

But I'm still not loving the drop-shadow ... Do we need that? Is it in the design?

image

src/components/PricingContracts/FacetSearch.svelte
matthampton-stock2shop: See https://github.com/stock2shop/console_v2/issues/128

src/components/PricingContracts/AddPricingContractModal.svelte
casperlotter: @matthampton-stock2shop This code deals with setting the value based on if you are editing or adding a new contract. If is a new contract its blank, otherwise we set it to the contract you are editing. The input is also then disabled if you are editing an existing contract.

matthampton-stock2shop: Ok - so is this kind of a round-about way of binding the Text input inside the widget with the search function to $contract.entityAttribute? Shouldn't we just do that directly rather? 

Doing it like this doesn't feel right to me ... 

Could you see if you can get it to work using the binding functionality of svelte instead? Or is there some reason not to do that?

src/stories/components/60-Search/FacetSearch.stories.svelte
```diff
@@ -5,14 +5,22 @@
...
   const staticFacets = $ConfigSystem.elasticSearch[index].facets;
   const selectedMetaFacets = $configPersistent.elasticSearch[index].metaFacets;
   const selectedFacetKeys = $configPersistent.elasticSearch[index].facetKeys;
-  const isDataInvalidated = writable(false);
-  const filterValue = writable('');
-  const data = writable(productJson as any);
+  const queryClient = new QueryClient();
+  const filterValue = '';
+  const doPostElasticSearch = (path: string, body: any, params: any): Promise<SystemProductsResponse> => {
+    const { subscribe } = writable(productJson as SystemProductsResponse);
```
matthampton-stock2shop, line 18: This looks a bit funny to me - don't think you should need to do anything like this...

Have a look at this example:

```
  async function doPostElasticSearch() {
    return (await orderJson) as SystemOrdersDataWrapper;
  }
  setPostElasticSearch(doPostElasticSearch);  <-- This sets the svelte-context so that you don't need to pass it as a prop to FacetSearch
```

https://github.com/stock2shop/console_v2/blob/69ef1893983decceac322ccf1798c744d7082b42/src/stories/routes/80-CustomerOrders.stories.svelte#L18

src/stories/components/60-Search/FacetSearch.stories.svelte
```diff
@@ -24,7 +32,9 @@
 />
 
 <Story name="Facet_Search">
-  <div class="h-full bg-white rounded overflow-y-auto min-w-[260px]  mr-6 shadow min-h-[900px]">
-    <FacetSearch {index} {staticFacets} {selectedMetaFacets} {selectedFacetKeys} {data} {isDataInvalidated} {filterValue} />
-  </div>
+  <QueryClientProvider client={queryClient}>
+    <div class="h-full bg-white rounded overflow-y-auto min-w-[260px]  mr-6 shadow min-h-[900px]">
+      <FacetSearch {index} {staticFacets} {selectedMetaFacets} {selectedFacetKeys} {doPostElasticSearch} {filterValue} />
```
matthampton-stock2shop, line 37: You shouldn't need `{doPostElasticSearch} ` here if you use `setPostElasticSearch` above...

src/components/PricingContracts/FacetSearch.svelte
```diff
@@ -5,15 +5,42 @@
...
 
   export let index: string;
   export let staticFacets: FacetWithIndex[];
   export let selectedMetaFacets: FacetWithIndex[];
   export let selectedFacetKeys: string[];
-  export let data: any;
-  export let isDataInvalidated: Readable<boolean>;
-  export let filterValue: Writable<string>;
+  export let filterValue: string;
+  export let doPostElasticSearch = usePostElasticSearch<SystemProductsResponse>();
```
matthampton-stock2shop, line 21: And this also need not be exported if you use `setPostElasticSearch` in the story

matthampton-stock2shop (changes requested):
Hi @casperlotter I think we're nearly there on this one! Just one more request...

matthampton-stock2shop (changes requested):
Hi @casperlotter I think we're nearly there on this one! Just one more request...

matthampton-stock2shop (approved):

PR #95, opened 2023-04-12
Justin krs/order fulfillment modal

justinengle-krs:
Fulfillment Modal

Zeplin:

image

Implementation:
image


casperlotter (dismissed):

matthampton-stock2shop:
https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=46

src/components/input/Dropdown.svelte
```diff
@@ -33,8 +33,8 @@
     {#if placeholder}
       <option value="" disabled>{placeholder}</option>
     {/if}
-    {#each options as option}
-      <option value={valueProperty ? option[valueProperty] : option} disabled={option.disabled ?? false}>
+    {#each options as option, i}
+      <option value={valueProperty ? option[valueProperty] : option} disabled={option.disabled ?? false} selected={!value && i == 0}>
```
matthampton-stock2shop, line 37: Do we always want the first item to display as selected if there isn't a value? (this feels like something that would be better to do in the place where the dropdown component is created (i.e. actually set the value to the correct default there)

src/models/orders/line-item.ts
```diff
@@ -120,5 +124,7 @@ export default class LineItem {
     this.total_display = total_display;
     this.price_display = price_display;
     this.total_discount_display = total_discount_display;
+    this.qty_already_fulfilled = this.fulfillments.reduce((sum, f) => sum + Number(f.qty), 0);
```
matthampton-stock2shop, line 127: These seem like view properties to me (or functions) - rather than editable model data properties... Thoughts?

src/models/orders/order-detail.ts
```diff
@@ -134,4 +135,19 @@ export default class OrderDetail {
     this.total_discount_display = total_discount_display;
     this.sources = sources;
   }
+
+  validateOrder(): { valid: boolean; error: string } {
```
matthampton-stock2shop, line 139: It looks to me like this function is specifically about fulfullments (rather than a general-purpose order validation function)... Would you mind renaming it to me make it a bit more specific?

matthampton-stock2shop (changes requested):

src/models/orders/line-item.ts
justinengle-krs: To me it makes sense for the model to own this logic?  Almost like a computed field?
I can almost see these coming from the API. The usage later seems intuitive as well. 

```const itemsToFulfill = systemOrder.line_items
      .filter((l) => l.qty_to_fulfill > 0)
      .map(
        (l) =>
          new FulfillmentOrderLineItem({
            price: l.price,
            qty: l.qty_to_fulfill,
            title: l.title,
            sku: l.sku,
          })
      );
```

would you prefer something like


```const itemsToFulfill = systemOrder.line_items
      .filter((l) =>getQtyToFulfill(l) > 0)
      .map(
        (l) =>
          new FulfillmentOrderLineItem({
            price: l.price,
            qty: l.qty_to_fulfill,
            title: l.title,
            sku: l.sku,
          })
      );
```

matthampton-stock2shop: I'd be fine for it to be a function on the model object (e.g. `l.getQtyAlreadyFullfilled()`) - I just don't really like it being a variable (that is just initialised to the right value - but could become wrong)

src/components/input/Dropdown.svelte
justinengle-krs: I thought it could be convenient default behavior of the control, if its set in the view that would still work

matthampton-stock2shop: But what if you want it to display a 'blank' state when the value is blank.. Can you do that somehow?

src/models/orders/line-item.ts
justinengle-krs: Are you happy if its a getter?


  get qtyAlreadyFullfilled() {
    return this.fulfillments.reduce((sum, f) => sum + Number(f.qty), 0);
  }

  get qtyToFulfill() {
    return this.qty - this.qtyAlreadyFullfilled;
  }

src/components/input/Dropdown.svelte
justinengle-krs: Thats true, let me make the change

src/models/orders/line-item.ts
matthampton-stock2shop: For sure!

src/models/orders/order-detail.ts
justinengle-krs: validateFulfillmentOrder ?

matthampton-stock2shop: Sounds fine... Thanks!

src/models/orders/line-item.ts
justinengle-krs: @matthampton-stock2shop  I forgot we do bind to qtyToFulfill on the modal, for each line item input.
Is that correct?

justinengle-krs: This field 

image

justinengle-krs: Which we then post on the 'Fulfill Order'. Which means this field at least would need to be a property on the model

matthampton-stock2shop: Does that make sense to you?

justinengle-krs: Yes, see latest commit. Happy with that?

matthampton-stock2shop (approved):

PR #104, opened 2023-04-21
User PUT

justinengle-krs:
Call PUT on user save changes without password


casperlotter (approved):

src/components/Settings/User.svelte
```diff
@@ -45,8 +45,24 @@
...
+      username: userVc.workingCopy.username,
+      active: userVc.workingCopy.active > 0,
+    });
+
+    await updateUser(userPut);
+  }
+
   async function saveNewPassword() {
-    loading = true;
     const userPut = new UserPut({
```
matthampton-stock2shop, line 66: I'm thinking that this should probably also check for userVc.isStale (since we're using the working copy here)...

matthampton-stock2shop (approved):

PR #109, opened 2023-04-26
Fix links from the orders and customers list screens to their detail screen, and the customer search result

matthampton-stock2shop:
This puts in the navigation action. It isn't perfect as it is just an onclick on the `tr` - it doesn't change the cursor to indicate clickability or the shading of the row on hover like the products one does.

However, these list pages haven't really been done yet by KRS - so I'm happy to leave that until they do. It's at least usable like this.


matthampton-stock2shop:
Hi @casperlotter Would you mind doing me a little code review?

matthampton-stock2shop:
https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=56

justinengle-krs (approved):

PR #111, opened 2023-04-26
Add support for customers that do not have orders

matthampton-stock2shop:
A customer may not have an associated user (e.g. it is created from a channel like shopify where the person involved is not ever going to be a stock2shop user).

In this case it should not show the user related fields:

- Price tier
- Warehouse
- Username
- Password stuff

image


matthampton-stock2shop:
This one is also ready for review, @casperlotter (chains on #109)

justinengle-krs (commented):
@matthampton-stock2shop  Was this the issue we had where the screen was just loading infinitely?

Issue #108, opened 2023-04-26
null / undefined should display as blank


matthampton-stock2shop:
Is there a general purpose way to handle this in svelte (e.g. React doesn't render anything for null/undefined)...

e.g. in this screen:

![image](https://user-images.githubusercontent.com/124143173/234561978-7a982109-ff18-4d56-b12f-bb040414671d.png)

TomRadford:
Solution is to render these optional properties with a conditional (`#if`) blocks. 
Have been employing this strategy to prevent undefined / null leaks into the UI as I work through the codebase.
@matthampton-stock2shop happy to close this?

matthampton-stock2shop:
> Solution is to render these optional properties with a conditional (`#if`) blocks. Have been employing this strategy to prevent undefined / null leaks into the UI as I work through the codebase. @matthampton-stock2shop happy to close this?

Sure... happy

PR #113, opened 2023-05-02
Handle a null instruction without a traceback

matthampton-stock2shop:
In this case the `getInstruction` function would return undefined, but then it would hit a traceback trying to read the label / description properties.

I've adjusted this not to fail. It does end up displaying 'undefined' which is a bit ugly - but see #108 - I want to know how to fix that in general rather than in each specific case, if there is a way.

I've also added a story to cover this case for future reference:

image

In the existing console it displays blank:

image


src/components/Orders/OrderHistory.svelte
```diff
@@ -28,7 +28,7 @@
   const instructionData = useQuery(['order-instruction-get', orderId], () => ordersRepo.getInstructions());
 
   function getInstruction(instruction: string) {
-    if (!instruction) return;
+    if (!instruction) return null;
```
matthampton-stock2shop, line 31: This is to stop lint from moaning at me about return types

matthampton-stock2shop:
See  https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=59

src/components/Orders/OrderHistory.svelte
justinengle-krs: I've not seen anything in svelte that changes the rendering of null/undefined behavior by default - I will maybe do some digging.

Should we have some global function perhaps that handles this and is used when rendering properties? Like a `formatString` or something like that.

Aurelia has value converters which was nice for this type of thing

justinengle-krs (approved):

PR #114, opened 2023-05-02
Customer without channel

matthampton-stock2shop:
This changes a couple of things:

1. The customer ID on the channel is moved down under the channel (it feels more logical there)
2. If there is not a channel selected yet, then there is not an error (attempting to access `systemChannel.type`)
3. The channel selection dialog stays enabled if there is no channel on the original record (so that you can select one, and then change if you accidentally clicked the wrong one). Once it is saved, you won't be able to any more.

image


matthampton-stock2shop:
See https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=60

justinengle-krs (approved):

Issue #117, opened 2023-05-03
Change of Role on the Settings->Users page does not save 


matthampton-stock2shop:
A refresh of the page shows that it did not save

PR #119, opened 2023-05-04
Interact with role user endpoint as needed

justinengle-krs:
Resolve https://github.com/stock2shop/console_v2/issues/117.

RoleUser Api was not being called


src/repos/role-users-repo.ts
```diff
@@ -27,6 +27,24 @@ export default class RoleUsersRepo {
     }
   }
 
+  async delete(userId: number): Promise<ApiResult<boolean>> {
```
matthampton-stock2shop, line 30: Is this the 'userId' or is it the role-user.id (i.e. the id of the row in the role-user table - rather than the user id? If so I think it should have a different name

src/components/Settings/User.svelte
```diff
@@ -36,6 +39,7 @@
   const roleUsersData = useQuery(['role-users', userId], () => roleUsersRepo.get(userId), {
     onSuccess(data) {
       const roleId = data.length ? data[0].role_id : 0;
+      roleUserId = data.length ? data[0].id : 0;
```
matthampton-stock2shop, line 42: I think I'm going to make some changes to this in #115 : I'd prefer the role-user ID to be stored with the roleIdVc object - so I'll do that and I think it will make this a bit simpler

src/repos/role-users-repo.ts
matthampton-stock2shop: (but I'm happy to change this in with my changes in #115)

matthampton-stock2shop (approved):

src/repos/role-users-repo.ts
justinengle-krs: roleUserId yes - my mistake, I thought it was userId and forgot to rename it.

PR #122, opened 2023-05-08
Settings product template add

laraDuPreezKrs:
In the list the existing product templates, user should be able to add new product template.
The user should not be able to save an empty product template and the only 2 fields required is the template name and attribute. 

Zeplin:
Capture

Implementation:
image


justinengle-krs (approved):

src/components/Settings/ProductAddTemplate.svelte
```diff
@@ -0,0 +1,90 @@
+<script lang="ts">
```
aStarlikeMine, line 1: I suggest renaming this file and the Row component to 'ProductTemplateAdd.

Reasoning is 'ProductTemplate' is the noun and 'Add' the verb, so it makes more sense to me to be ordered that way. 
It'll also then be listed with the ProductTemplate component in the Settings file should other product files with other names be added.

src/components/Settings/ProductAddTemplate.svelte
```diff
@@ -0,0 +1,90 @@
...
+</script>
+
+<CenterModal bind:open>
+  <div class="flex justify-between h-auto w-full" slot="title">
+    <div class="flex items-center gap-4">
+      <Btn Prefix={IconBack} size="small" iconSize="small" variant="gray" onClick={reset}/>
+      <div class="flex items-center"><TextInput bind:value={templateName} className="w-96" /></div>
+    </div>
+    <div class="flex items-center">
+      <Btn onClick={saveTemplate} size="medium" variant="solid">Save changes</Btn>
```
aStarlikeMine, line 69: Should the modal stay open on save or should it close and use the list notification?

aStarlikeMine (approved):

src/components/Settings/ProductTemplate.svelte
```diff
@@ -0,0 +1,93 @@
...
+
+    if(hasData.length && templateName){
+      const productTemplatesPostResult = await Promise.all(
+        templateItems.filter((item) => item.key).map(async(templateItem: TemplateItem) => {
+          templateItem.name = `product_${templateName}`;
+          return await productTemplateRepo.post(templateItem);
+        })
+      )
+      inlineNotification = {
+        type: productTemplatesPostResult.at(-1).result ? 'success' : 'error',
```
matthampton-stock2shop, line 42: I think we should check if any of the saves failed here

src/components/Settings/ProductTemplate.svelte
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+  }
+
+  function reset(): void{
+    open = false;
+    templateItems = [new TemplateItem({ type: 'String'})];
+    templateName = '';
+    displayInlineNotification = null;
+    const queryClient = useQueryClient();
```
matthampton-stock2shop, line 60: This useQueryClient should probably be outside the function rather - not sure how it behaves when called inside another function (might be fine - but looks a little outside the norm to me)

src/models/product-templates/system-meta-template-item.ts
```diff
@@ -0,0 +1,29 @@
+export default class SystemMetaTemplateItem {
```
matthampton-stock2shop, line 1: Things feels like a bit of duplication over the src/models/product-templates/meta-template-item.ts above... Are we sure we need both?

src/components/Settings/ProductTemplateRow.svelte
```diff
@@ -0,0 +1,35 @@
...
+<div>
+  <div class="items-center grid grid-cols-12 gap-2 xl:gap-5 pb-2 pt-4 mx-6 border-b border-neutral-200">
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-3">Attribute</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-2">Type</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-3">Default</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-3">Description</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller place-items-start col-span-1">Required</div>
+  </div>
+
+  {#each templateItems as templateItem, i}
```
matthampton-stock2shop, line 23: I'm thinking that this component should rather be called `ProductTemplateRows` since it seems to be for all of the rows, or `ProductTemplateItems` maybe?

src/components/Settings/ProductTemplateRow.svelte
```diff
@@ -0,0 +1,35 @@
...
+    <div class="items-center grid grid-cols-12 gap-2 xl:gap-5 py-4 mx-6 border-b border-neutral-200 ">
+      <TextInput bind:value={templateItem.key} className="col-span-3 flex-col" />
+      <Dropdown options={options} bind:value={templateItem.type} className={'col-span-2 flex-col'} display_func={(o) => o} />
+      <TextInput bind:value={templateItem.default} className="col-span-3 flex-col" />
+      <TextInput  bind:value={templateItem.description} className="col-span-3 flex-col" />
+      <div class="flex items-center place-content-start">
+        <Toggle bind:checked={templateItem.required} className="border-none px-0 mr-1 xl:mr-4" />
+        <Btn onClick={() => deleteTemplateItem(i)} size="small" variant="gray" Prefix={IconDelete} iconSize="extra-small" class="hover:!bg-neutral-100 !text-neutral-700" />
```
matthampton-stock2shop, line 31: I think it's possible, even if unlikely, to have some unfortunate race conditions if we use the index of the item in the templateitems list as the parameter for the delete function. I'd prefer to pass the id along here - just to be sure... 

Although, I guess it may not have an ID yet? (if this is an add)

But I'm assuming we'll re-use this component for the edit interaction as well...

Maybe we can give the items a temporary id when we add it that could just be the index, and gets replaced with the proper server-assigned ID for the edit interaction?

[This could also be done when the edit is implemented]

Am I making sense? Or overcomplicating things?

src/components/Settings/ProductTemplate.svelte
laraDuPreezKrs: Have a look at my changes, I've taken into account when there's only successful results or only errors and then when there's results and errors (I used 'info' here) the 400 bad request messages don't look that nice it's a bit long and not user friendly. Let me know what you prefer

image

src/components/Settings/ProductTemplateRow.svelte
laraDuPreezKrs: I'm leaning towards `ProductTemplateItems` . @aStarlikeMine your reusing this for Edit, what do you think?

src/models/product-templates/system-meta-template-item.ts
laraDuPreezKrs: Only needed the system-meta-template-item model. Thanks

src/components/Settings/ProductTemplateRow.svelte
```diff
@@ -0,0 +1,35 @@
...
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-2">Type</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-3">Default</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller col-span-3">Description</div>
+    <div class="flex items-center text-neutral-500 font-bold text-smaller place-items-start col-span-1">Required</div>
+  </div>
+
+  {#each templateItems as templateItem, i}
+    <div class="items-center grid grid-cols-12 gap-2 xl:gap-5 py-4 mx-6 border-b border-neutral-200 ">
+      <TextInput bind:value={templateItem.key} className="col-span-3 flex-col" />
+      <Dropdown options={options} bind:value={templateItem.type} className={'col-span-2 flex-col'} display_func={(o) => o} />
```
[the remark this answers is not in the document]
laraDuPreezKrs: The dropdown component allows for and array of any, the export is called options.
 I'll remove the identity function from this dropdown component here, but would leave it in the dropdown component as it's used when the options array is an object so we can specify which property to display. Let me know if your happy with that.

src/components/Settings/ProductTemplate.svelte
aStarlikeMine: @laraDuPreezKrs what about, in the case of a mixed result, 'error' instead of 'info' is used and the message changes to something like 'Successfully added item however some attribute(s) had an error' and then concatenate them as you have done.

src/components/Settings/ProductTemplateRow.svelte
aStarlikeMine: Agree on the plural. Items or Rows, whichever is more commonly used in the project

src/components/Settings/ProductTemplate.svelte
laraDuPreezKrs: Thanks for the suggestion @aStarlikeMine, I will definitely use your wording. It's much better than my generic wording.  
I'm happy to do what @matthampton-stock2shop think is suitable for the project.

matthampton-stock2shop (approved):

matthampton-stock2shop (approved):

PR #123, opened 2023-05-08
Delete a Product Template

aStarlikeMine:
Allow the deletion of a product template from the list page in Settings > Product Templates.

_Note:_
- the notification doesn't disappear automatically. This is current behaviour with the InlineNotification component. I'm reluctant to change it without discussion on design and project requirements. 

Successful deletion:
image

Unsuccessful deletion _(if there is no error message from the api)_ :
image


justinengle-krs (approved):

laraDuPreezKrs (commented):
On save the inline notification is red.
image

matthampton-stock2shop (approved):

PR #135, opened 2023-05-22
Justin krs/customer add

justinengle-krs:
Implentation:
image

Zeplin:
image


casperlotter (dismissed):

src/components/Customers/AddCustomerModal.svelte
```diff
@@ -0,0 +1,194 @@
...
+
+<SideModal bind:open saveText="Add Customer" saveClicked={addCustomer}>
+  <div class="flex" slot="title"><IconCustomer /> <span class="ml-[16px]">Add Customer</span></div>
+  <div slot="content">
+    {#if isLoading}
+      <Loader />
+    {:else}
+      <div class="px-8 py-8 border-b border-neutral-200">
+        <div class="text-regular text-brand-action font-bold mb-[16px]">DETAILS</div>
+        <TextInput title="Company Name" bind:value={customer.companyName} className="mt-[16px] w-full" />
```
matthampton-stock2shop, line 146: This company name isn't being used, I don't think (since it goes in to the first address - which is odd) - I think let's hide it from the dialog for now (and actually from the details screen - where it looks like you can edit it).

Also fine to do in a follow-up. See https://github.com/stock2shop/console_v2/issues/163

src/models/channel-customers/channel-user.ts
```diff
@@ -0,0 +1,14 @@
+export interface ChannelUser {
+  id: number;
+  created: string;
+  admin: number;
+  name: string;
+  surname: string;
+  email: string;
+  active: number;
+  salt: string;
+  password: string;
```
matthampton-stock2shop, line 10: Our existing console asks you to supply a password for a new customer - which we aren't doing now (which I am fine with) - but I also can't then set a password afterwards. See https://github.com/stock2shop/console_v2/issues/164. 

Happy for this to be fixed as a follow-up

PR #142, opened 2023-05-29
Casper krs/product details options

casperlotter:
WIP


justinengle-krs (changes requested):
Comment on here
https://www.chromatic.com/test?appId=63ef6cddab1dec384359e1cd&id=64818aceb82ad46fae8e741d

src/components/Customer/CustomerDetail.svelte
```diff
@@ -64,14 +62,10 @@
   const channelData = useQuery(['channels-get'], () => channelsRepo.get());
 
   const userData = useQuery(['user-get', userId], () => usersRepo.get(userId), { enabled: !!userId });
-
   $: user = $userData.data || null;
   $: userId = $customerData.data?.user?.user_id;
```
casperlotter, line 66: @justinengle-krs I reverted this change to reference the user_id. But it is definitely broken. I suspect the data is being cast to a SystemCustomerUser which defaults the user_id to 0. So I suspect that this will always be the case since the data the customer brings back is a User that does not have a user_id field

Ayano2000 (commented):
@matthampton-stock2shop 
I've clicked through the changes and everything works for me!
My only question is that when I click on the channels tab I only see a message that looks like `Channels for product <product_id>` I'm assuming this is still to be implemented?
Storybook looks good as well.

As for the code changes I haven't noticed anything that looks fishy, but it's possible I might have missed something..

src/components/Product/ProductOption.svelte
```diff
@@ -1,88 +1,109 @@
...
+  export let addItem: (item: ProductVariantEntity) => void = () => {};
+  export let removeItem: (item: ProductVariantEntity) => void = () => {};
+
+  let showAddItem = false;
+
+  let container: HTMLElement;
+  const toggleAddItem = () => (showAddItem = true);
+  const hideAddItem = () => (showAddItem = false);
+
+  const onWindowClick = (e: MouseEvent) => {
```
matthampton-stock2shop, line 22: I know this was already here - but it seems a bit low-level... Do we need this?

matthampton-stock2shop (commented):
Hi @casperlotter ... I've not done a very thorough review at all - @justinengle-krs could you look more carefully for me?

I did make a few suggestions - specifically relating to the use of the bind functionality of svelte - and the fact that the versioncontrolled classes are svelte 'stores'.

It might be worth reading up a bit about how those work - I think it could simplify some of this code a bit maybe

src/components/Product/ProductDetailsConfigurator.svelte
```diff
@@ -1,105 +1,185 @@
...
+  async function saveChanges(): Promise<void> {
+    const valid = variant.validate();
+
+    if (!valid || noChangesMade) {
+      return;
+    }
+    let verb = variant.newVariant ? 'added' : 'updated';
+    let result = variant.newVariant ? await repo.post(variant.toVariant()) : await repo.put(variant.toVariant());
+
+    const success = result.result ? true : false;
```
justinengle-krs, line 43: seems unecesarry to check if a boolean is true or false and then return true or false

justinengle-krs (changes requested):
I can add a null warehouse and price list

image
image

justinengle-krs (changes requested):
image
Can you only add 3 options?
Also in the zeplin design The tick is green, and theres a state where its a locked icon, not sure what thats for

justinengle-krs (changes requested):
Is this supposed to do something or is it a separate card?
image

src/components/Product/ProductDetailsConfigurator.svelte
casperlotter: @justinengle-krs Result is an ApiResult in this context:

`export type ApiResult<T> = {
  result: T;
  error: string | undefined;
};
`

src/components/Product/ProductVariants.svelte
```diff
@@ -2,17 +2,26 @@
...
-  // Modal
-  let showModal = false;
-  const toggleModal = () => { setTimeout(() => { showModal = !showModal; }, 0); };
-  const hideModal = () => { showModal = false; };
+  let sendUpdateMessage = (event) => {
+    dispatch('updated', event.detail);
+  };
+
+  export let variants: ProductVariant[] = [];
+  export let expandedId = variants.length > 0 ? variants[0].id : 0;
```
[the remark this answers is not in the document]
casperlotter: Not sure what you are asking here @justinengle-krs

casperlotter:
> image Get this on save as well

@mannyqi The variant error I raised in Keybase since it seemed to be an API issue but the comments around it have been exploded. Can you please assist?

src/components/Product/ProductDetailsConfigurator.svelte
justinengle-krs: result.result is a boolean though

So it's going const success = <boolean> ? <boolean> : <boolean>
Where result.result just has that value already.

src/components/Product/ProductVariants.svelte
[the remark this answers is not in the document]
justinengle-krs: Shouldn't need the > 0

variants.length ? variants[0].id : 0;

mannyqi:
@casperlotter I can't seem to reproduce the issue on my side. Can you send me the exact steps please?

casperlotter:
@mannyqi you can navigate to #/product/3163058 and attempt to save an option against one of the variants.

mannyqi:
@casperlotter I managed to reproduce the issue, thanks.

While doing so I picked up something else. The "Save changes" button would not activate up until after I added a warehouse, but there was nothing to indicate why I was unable to save. I only figured it out after digging through the code a little...Not sure if validation for require fields will come later?

image

casperlotter:
@mannyqi I added the errors as part of the inline notification now. If you have a variant open it will display the validation errors for that variant.

mannyqi:
@casperlotter I thought I reproduced the error you mentioned, but it was actually a different error:

```
Bad request - Product option name 2 missing
```

console-v2-options-error

I suspect the error you are getting is specific to the product you are looking at.
What happens when you do the same thing on a different product?

--- 

### Old console workflow for adding new option:

I looked at the old console's workflow for this feature and it is as follow:

1. Edit product and add new option name under the "General" tab
```
PUT http://stock2shop/v1/products/{product_id}?format=json&token={token}
```
new-option-1

2. Click "Save"
3. Then go to the "Variants" tab and enter a new option name under a specific variant and click the "Save" button for that variant
```
PUT /v1/products/variants/{variant_id}?format=json&token={token}
```
image

mannyqi:
## Styling issues:

1. Option name icon should be green and notice icon should be triangle
image


2. Missing "Grid Edit" feature?
image

3. When saving a variant, there is no "success" message. "Save changes" button just disables and page moves to the top. Not sure if this is by design?
4. Spacing between variants are 8px, instead of 12px
image

5. Some of the empty variant Details' labels are gray, whereas the entered ones are black. Not sure if the idea here is that the labels only goes black after information has been entered?
image

mannyqi (changes requested):
@casperlotter I'm happy. just a couple of small changes.

- "Barcode" should not be a required field

mannyqi:
@casperlotter all working, thanks

justinengle-krs (approved):

PR #143, opened 2023-05-29
Fix chromatic build error introduced in #141

matthampton-stock2shop:
That can be an empty list - so -1 would be the result. Fixes error shown in https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=120.

This was broken in a9f932485d105e8540474c47fe039914feba7b12 - merged in #141


matthampton-stock2shop:
Chromatic: https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=121

justinengle-krs:
> That can be an empty list - so -1 would be the result. Fixes error shown in https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=120.
> 
> This was broken in [a9f9324](https://github.com/stock2shop/console_v2/commit/a9f932485d105e8540474c47fe039914feba7b12) - merged in #141

Would probably affect customersgrid and ordersgrid as well? as well as the grid headers?

justinengle-krs (approved):

PR #144, opened 2023-05-29
Minor tweaks for demo

matthampton-stock2shop:
- Navigate to the dashboard from the home URL
- Stop the customers -> orders screen from being full width
- Make the b2b url configurable
- Update the app title and favicon


src/layouts/Console.svelte
```diff
@@ -8,7 +8,7 @@
 
   $: isSettings = $location.includes('/settings');
   $: isFullWidth =
-    $location.includes('/dashboard') || $location.endsWith('/products') || $location.endsWith('/orders') || $location.endsWith('/customers');
+    $location.startsWith('/dashboard') || $location.startsWith('/products') || $location.startsWith('/orders') || $location.startsWith('/customers');
```
justinengle-krs, line 11: why this change? Just curious

I was a bit worried about potential routes would be /products/something
But I guess its not going to be a thing

matthampton-stock2shop: Then reason was the inverse of that worry (that is actually a thing): `http://localhost:3003/#/customer/6263834/orders` was going full width cos it endsWith('/orders')

justinengle-krs (approved):

PR #156, opened 2023-06-08
Implement product master component for dynamic pricing

casperlotter:
Zeplin:
Screenshot 2023-06-08 111120

Implementation:
Screenshot 2023-06-08 111134

https://www.chromatic.com/build?appId=63ef6cddab1dec384359e1cd&number=131


justinengle-krs (approved):

mannyqi:
@casperlotter I got the following errors in my console when viewing the "Dynamic Pricing" tab

image

After adding a product meta for product ID 3171421 the page loaded without error.

Screenshot 2023-06-14 at 09 53 23

A few other things I picked up:
1. Left and right padding on "Dynamic Pricing" tab is less than that in zeplin - or rather the tab width is less.
2. PageHeader and ProductNav components are [imported](https://github.com/stock2shop/console_v2/blob/ecbb78429c183daa47b6233fe767fb11e3156d81/src/routes/ProductContract.svelte#L2-L3) but not in use
3. The contract deletes with no confirmation

casperlotter:
@mannyqi Updated. I am not able to reproduce your error. If I open that product I get an issue because it is not linked to my current logged in user but if I operate on products in the scope of this account everything works fine.

mannyqi:
Thanks @casperlotter 

I get the error in the console for any product that has no product meta. Can you confirm whether the product you viewed had a meta or not? I'll see if I can ask on of the other guys to check if they have this issue as well.

Then last couple of things:

#### 1. Tab page heading
There seems to be a difference in the left & right padding for the tab page heading causing the heading & button to not line up correctly with the rest of the content below it.

Notice below the padding for one of the other tab headings which is aligned correctly:

Screenshot 2023-06-19 at 08 31 33

But then for dynamic pricing, the padding is a lot more.

image

--- 

#### 2. The tabs
The active and hover padding for the "Dynamic Pricing" tab are different. This seems to be causing the tabs to jump around when navigating between them. e.g. try navigating between "Attributes" and "Data Sync". The tabs all stay in place. Then try navigating between "Data Sync" and "Dynamic Pricing". The tabs will then move around.

casperlotter:
@mannyqi I updated the padding for the content. The current tabs should not be compared with the other pages since they have not been built yet and need to be updated.

I added in an inline notification that should display now if an issue occurs with the product not having any data.

mannyqi:
> The current tabs should not be compared with the other pages since they have not been built yet and need to be updated.

Noted

> I added in an inline notification that should display now if an issue occurs with the product not having any data

The issue still remains - getting a js error in developer console for products with no meta.

For some reason the 400 error is not being handled through the "onError" function? 
Even causing a 500 error still comes through as a successful request handled by the "onSuccess" function.

Does this work on your side?

src/routes/ProductContract.svelte
```diff
@@ -74,28 +86,27 @@
...
-  <div>
-    <PageHeader title="Cotton T-Shirt" hasBackButton={true}>
-      <Btn size="large" variant="solid" disabled={true}>Save changes</Btn>
-    </PageHeader>
-    <ProductNav {productLinks} />
+  <ProductMaster selectedTab="dynamic-pricing" {productVersionControlled} {productsRepo}>
     <DynamicPricing bind:open {addContractModel} metaRepo={productMetaRepo} productId={params.id} {inlineNotification} {invalidateCache} />
     <div class="overflow-hidden">
-      <div class="grid grid-cols-[1fr,3fr,3fr,1fr,1fr,1fr] w-9/10 ml-12 mr-12 gap-x-8 border-b border-neutral-200 items-center py-3 px-6 leading-4">
+      <div class="grid grid-cols-[1fr,3fr,3fr,1fr,1fr,1fr] w-9/10 ml-12 mr-12 gap-x-8 border-b border-neutral-200 items-center py-3 leading-4">
```
mannyqi, line 111: Not sure why the `px-6` class was removed as the list headers indentation is now gone.
Also, the padding on both the left & right of the list headers and data should be 16px (px-4) as per zeplin:

image

mannyqi: I also noticed that the tab page heading and "Add Contract" button is not aligning with the lists outer edges.

The list has left & right margins of 48px while the page heading has left & right padding of 40px. As per zeplin design, this should be 32px:

image

casperlotter:
@mannyqi Yes, we need to look at how we deal with errors in our repo calls so that we can use the native usequery callbacks instead of the way we are doing them now. I updated the code to do it the way we are handling it elsewhere and it works well for me now.

mannyqi (changes requested):

mannyqi:
@casperlotter the error handling is working for products with no meta, thanks.

Please have a look at the css issues I commented on.

casperlotter:
@mannyqi I updated the styling.

mannyqi:
I picked up another issue - the list does not refresh after removing last item on the second page for example. So even though there are more items you end up with an empty list. See screens below:

Screenshot 2023-07-05 at 13 45 26

Screenshot 2023-07-05 at 13 46 39

PR #174, opened 2023-06-29
Justin krs/images import progress

justinengle-krs:
Implements the matching of images to the product SKUS:

image

Zeplin:
image


casperlotter (approved):

mannyqi:
## Styling & Functionality
1. "Add Rule" button uses incorrect font-size and padding (or width & height)
image

2. The "Remove All" and "Upload All" buttons are small than in design
image

3. There should be `16px` spacing between buttons
image

4. Table headers should be `11px` and not `13px`
image

5. Each row should have `24px` padding to the left and right
image

6. When uploading a single image exceeding the max upload size (5mb), then an appropriate error message is displayed. However, when uploading multiple images at the same time, then large images are simply ignored (no error message is displayed) and the rest is added to the queue. Perhaps these files should still be included in a list but with a status of "Too Large" or maybe instead of having the message "Failed to match product, image ignored!", have "File size is too large. Max 5MB allowed"?
image

7. Blue text in top-left corner should be `13px`, not `14px`
image

8. In the zeplin design it looks like the remove and upload buttons are immediately below the list of uploaded files. So I would assume the buttons are pushed down as the list grows. But I think always having it visible at the bottom (like you have it) might be better, but it just needs a white background.
image

mannyqi:
One that I missed:

9. Unsupported file format error message is not styled properly
image

src/components/Import/ImportImagesModal.svelte
```diff
@@ -1,119 +1,282 @@
...
+    const fieldName = selectedField.startsWith('variant.') ? selectedField.replace('variant.', '') : selectedField.replace('product.', '');
+
+    const isMatch = (product: SystemProduct) => {
+      const fieldValue = selectedField.startsWith('variant.')
+        ? product.variants?.some((variant: any) => variant[fieldName]?.toLowerCase() === searchTerm.toLowerCase())
+        : product[fieldName]?.toLowerCase() === searchTerm.toLowerCase();
+      return Boolean(fieldValue);
+    };
+
+    const matchedProduct = products.find(isMatch);
```
mannyqi, line 129: We can just return the value here instead of creating const

src/components/Import/ImportImagesModal.svelte
```diff
@@ -1,119 +1,282 @@
...
+    return matchedProduct;
+  }
 
-  let findReplaceRules = [];
-  function addFindReplaceRule() {
-    const newRule = {
-      find: findValue,
-      replace: replaceValue,
+  function getRequestBody(query: string[]): any {
+    const esSearch = {
```
mannyqi, line 134: We can just return the value

justinengle-krs:
> ## Styling & Functionality
> 1. "Add Rule" button uses incorrect font-size and padding (or width & height)
> 
> image
> 2. The "Remove All" and "Upload All" buttons are small than in design
> 
> image
> 3. There should be `16px` spacing between buttons
> 
> image
> 4. Table headers should be `11px` and not `13px`
> 
> image
> 5. Each row should have `24px` padding to the left and right
> 
> image
> 6. When uploading a single image exceeding the max upload size (5mb), then an appropriate error message is displayed. However, when uploading multiple images at the same time, then large images are simply ignored (no error message is displayed) and the rest is added to the queue. Perhaps these files should still be included in a list but with a status of "Too Large" or maybe instead of having the message "Failed to match product, image ignored!", have "File size is too large. Max 5MB allowed"?
>    image
> 7. Blue text in top-left corner should be `13px`, not `14px`
> 
> image
> 8. In the zeplin design it looks like the remove and upload buttons are immediately below the list of uploaded files. So I would assume the buttons are pushed down as the list grows. But I think always having it visible at the bottom (like you have it) might be better, but it just needs a white background.
>    image

Regarding some of the buttons -  I am hesitant to change this as the project has predefined and consistent button variants and styles, I don't think we should implement  zeplin pixel perfects if it doesn't align with the other screens? Unless something is way off then we'd need a new variant

The same with the font sizes the app consistently uses the following sizes, and I think we should stick to whichever is the closest, if its within 1px?

```
        smaller: '0.625rem',
        small: '0.688rem',
        regular: '0.813rem',
        big: '1.25rem',
```

Let me know what you think? 
Padding and mis-alignments and things that look off I will resolve though as well as  the functionality issues

mannyqi (changes requested):
Can we re-apply the rules when switching between the fields as well?

mannyqi:
> I am hesitant to change this as the project has predefined and consistent button variants and styles, I don't think we should implement zeplin pixel perfects if it doesn't align with the other screens?
> 
> The same with the font sizes the app consistently uses the following sizes, and I think we should stick to whichever is the closest, if its within 1px?

I was instructed to also check styling, so can't say. Let me try and check with Chris first and get back to you.

justinengle-krs:
> > I am hesitant to change this as the project has predefined and consistent button variants and styles, I don't think we should implement zeplin pixel perfects if it doesn't align with the other screens?
> > The same with the font sizes the app consistently uses the following sizes, and I think we should stick to whichever is the closest, if its within 1px?
> 
> I was instructed to also check styling, so can't say. Let me try and check with Chris first and get back to you.

Okay no problem. I will get everything else sorted in the meantime

src/components/Import/ImportImagesModal.svelte
```diff
@@ -1,119 +1,282 @@
 <script lang="ts">
-  import type { StringRange } from '@/lib/es/types';
   import Icon from '../Icons/Icon.svelte';
```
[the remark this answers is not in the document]
justinengle-krs: Done

src/components/Import/ImportImagesModal.svelte
justinengle-krs: Done

src/components/Import/ImportImagesModal.svelte
justinengle-krs: Done

justinengle-krs:
> > I am hesitant to change this as the project has predefined and consistent button variants and styles, I don't think we should implement zeplin pixel perfects if it doesn't align with the other screens?
> > The same with the font sizes the app consistently uses the following sizes, and I think we should stick to whichever is the closest, if its within 1px?
> 
> I was instructed to also check styling, so can't say. Let me try and check with Chris first and get back to you.

@mannyqi Points 3,5,6,8,9 have been resolved. The table headers are 11px

mannyqi:
@justinengle-krs happy with all the changes, thanks.

Just one final request - just like in point 6, when uploading a single unsupported file you get the error message like below. I think this should work the same as the others - i.e. unsupported files should still be listed in the queue, but with an appropriate message and status (e.g. Unsupported)

image

justinengle-krs:
> @justinengle-krs happy with all the changes, thanks.
> 
> Just one final request - just like in point 6, when uploading a single unsupported file you get the error message like below. I think this should work the same as the others - i.e. unsupported files should still be listed in the queue, but with an appropriate message and status (e.g. Unsupported)
> 
> image

@mannyqi Change has been made

mannyqi:
Thanks @justinengle-krs 
Happy with the change.

mannyqi:
@justinengle-krs the `Product/DataSync` story is failing to load.

image

justinengle-krs:
> @justinengle-krs the `Product/DataSync` story is failing to load.
> 
> image

on it

justinengle-krs:
> > @justinengle-krs the `Product/DataSync` story is failing to load.
> > image
> 
> on it

Fixed @mannyqi

justinengle-krs:
> @justinengle-krs
> 
> 1. After finishing the uploading of images, you end up with the screen below. When you click the back button in the top-left corner, it closes the modal. If you click "Import Images" again, it opens the same window (after completing upload - 2nd screen) instead of the first step allowing you to upload images.
>    image
> 
> image

@mannyqi This has been resolved

mannyqi (approved):

PR #179, opened 2023-07-04
Bug When editing a product the source_variant_code is-populated-into-the-product-code-field-

justinengle-krs:
Seems to be just an incorrect label


src/models/products/product-variant.ts
```diff
@@ -112,11 +112,11 @@ export default class ProductVariant {
       unit: 'grams',
     });
     details = details.concat({
-      name: 'Product Code',
+      name: 'Source Variant code',
```
mannyqi, line 115: Please change name to "Variant Code"

mannyqi (changes requested):

PR #180, opened 2023-07-04
Bug: Console v1 tags are loaded as a single tag

justinengle-krs:
use correct delimiter and more robust fitlering


casperlotter (dismissed):

mannyqi:
After the initial load of the tags, the select still lists all the already populated tags (they are not selectable though)
image

justinengle-krs:
> After the initial load of the tags, the select still lists all the already populated tags (they are not selectable though) image

Resolved

mannyqi (approved):

PR #190, opened 2023-07-12
Allow for new tags to be added

casperlotter:
Screenshot 2023-07-12 134543


justinengle-krs (changes requested):

justinengle-krs (approved):

src/components/form/MultiSelect.svelte
```diff
@@ -4,25 +4,56 @@
...
 </script>
 
-<Select value={initVal} {items} {isMulti} on:input={update} on:change={update} on:clear={update} />
+{#if hasItems}
+  <Select bind:value={currentValues} {items} {isMulti} on:input={update} on:change={update} on:clear={update} />
+{/if}
+
+<input
+  type="text"
+  class="h-10 w-full rounded border border-neutral-200 p-3 text-regular focus:shadow-none focus:border-neutral-200 focus:ring-0 text-neutral-700 placeholder-neutral-400 s-q19PVNo4u0FE border-dashed"
```
matthampton-stock2shop, line 55: What is 's-q19PVNo4u0FE'?

src/components/form/MultiSelect.svelte
```diff
@@ -4,25 +4,56 @@
...
+  bind:value={new_tag}
+  on:keydown={newTagInputChanged}
+  placeholder="Add Tag"
```
matthampton-stock2shop, line 58: Is the "Multiselect" component only used for Tags? Maybe rename to "Tags" component, if so?

matthampton-stock2shop: (or this needs to be an arg)

matthampton-stock2shop (changes requested):

matthampton-stock2shop:
@justinengle-krs Isn't there a component already that lets you add new items to a 'comma-separated' list? Do we need this?

casperlotter:
> @justinengle-krs Isn't there a component already that lets you add new items to a 'comma-separated' list? Do we need this?

@matthampton-stock2shop The CommaSeparateInput has no concept of available values that is not assigned like we need for product tags. Unless there is a misalignment on our understanding of what the functionality should be here.

Currently we fetch all product tags on a separate call from [GET](https://app.stock2shop.com/docs/#!/products/getMetaData_get_2) [/products/meta](https://app.stock2shop.com/docs/#!/products/getMetaData_get_2) that is then used as options to be selectable as a dropdown for the product if it is not already assigned to the specific product. The Tags component gives visibility on what existing tags are available while editing. With the CommaSeparatedInput component this is not possible and only assigned tags to the specific product can be seen initially and not when cleared.

matthampton-stock2shop:
> > @justinengle-krs Isn't there a component already that lets you add new items to a 'comma-separated' list? Do we need this?
> 
> @matthampton-stock2shop The CommaSeparateInput has no concept of available values that is not assigned like we need for product tags. Unless there is a misalignment on our understanding of what the functionality should be here.
> 
> Currently we fetch all product tags on a separate call from [GET](https://app.stock2shop.com/docs/#!/products/getMetaData_get_2) [/products/meta](https://app.stock2shop.com/docs/#!/products/getMetaData_get_2) that is then used as options to be selectable as a dropdown for the product if it is not already assigned to the specific product. The Tags component gives visibility on what existing tags are available while editing. With the CommaSeparatedInput component this is not possible and only assigned tags to the specific product can be seen initially and not when cleared.

Ah .. ok

matthampton-stock2shop (approved):

PR #191, opened 2023-07-17
Justin krs/show all settings

justinengle-krs:


src/components/Connectors/ChannelMaster.svelte
```diff
@@ -3,7 +3,6 @@
   import PageHeader from '@/components/PageHeader.svelte';
```
matthampton-stock2shop, line 3: RE  image

I think I'd rather see them expanded all the time (i.e. maybe no need for a collapsible thing)

matthampton-stock2shop, line 3: RE
image

It feels like the title and "help text" are back to front here

matthampton-stock2shop, line 3: Also, rE
image

When there are fixed options like this, I'd like to see a drop-down ... Does the meta template give you the options?

Blank should be in the option list

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -30,29 +41,162 @@
       navigation: 'info-display',
       label: 'Info Display',
       IconComponent: IconInfo,
-      fields: ['account_display', 'checkout_fields', 'order_columns', 'order_view_display', 'payment_methods'],
+      fields: [
+        'account_display',
```
matthampton-stock2shop, line 45: Are the options displayed in this order? (i.e. there is a way to specify the order)

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -30,29 +41,162 @@
...
+        'order_columns',
+        'order_view_display',
+        'account_invoice',
+        'account_statement',
+        'address_line1',
+        'address_line2',
+        'address_line3',
+        'channel_order_code_prefix',
+        'channel_order_code_sequence',
+        'currency',
```
matthampton-stock2shop, line 55: I think we should add a new nav category for 'Currency and Tax'

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -30,29 +41,162 @@
...
+        'display_name',
+        'elastic_query_fields',
+        'elastic_suggest_fields',
+        'hide_tax',
+        'hmac_shared_secret',
```
matthampton-stock2shop, line 60: And a "Other Settings" bag where we can put odd bits and bobs like this

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop, line 3: RE
image

I think these 'generic' KeyValue config things are a bit confusing

I'd like to see 'props' to give a specific title to use for 'Key' and for 'Value' and a specific placeholder text instead of 'KEY_0')

e.g. For "Option Sort Order" I'd like to see "Option" instead of "Key", and 'e.g. Size' instead of "KEY_0", and I'd like to see "Sort Order" instead of "Value", and also have placeholder for the value there of "e.g. XL,L,M,S,XS"

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -3,7 +3,6 @@
   import PageHeader from '@/components/PageHeader.svelte';
   import Sidebar from '@/components/Connectors/Sidebar.svelte';
```
matthampton-stock2shop, line 4: RE image

These look like they should be toggles ...

Is the meta template not giving enough information? (I see that there are some toggle ones in other places)

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -30,29 +41,162 @@
...
+        'currency',
+        'display_name',
+        'elastic_query_fields',
+        'elastic_suggest_fields',
+        'hide_tax',
+        'hmac_shared_secret',
+        'industry',
+        'logo',
+        'option_sort_order',
+        'over_order_enabled',
```
matthampton-stock2shop, line 64: I also think we need another category for "Order Settings"

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -30,29 +41,162 @@
...
+        'phone',
```
matthampton-stock2shop, line 65: image

This should be a pick-list of fulfilment services

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop, line 4: image

I think I'd also be inclined to just have these always expanded

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -3,7 +3,6 @@
   import PageHeader from '@/components/PageHeader.svelte';
   import Sidebar from '@/components/Connectors/Sidebar.svelte';
   import ChannelsRepo from '@/repos/channels-repo';
```
matthampton-stock2shop, line 5: image

This screen gets very confused if you add the same payment method twice! Probs shouldn't allow it...

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -3,7 +3,6 @@
...
-  import { onMount } from 'svelte';
```
matthampton-stock2shop, line 6: image

Some slightly funky padding / overlap here

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: Or maybe move the currency and tax related items under Pricing for now

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Okay no problem. @mannyqi suggested the collapsible because of how long the list may be. Should I remove?

mannyqi: @matthampton-stock2shop sorry, this was my suggestion as the list of configurations was very long and confusing to look at...

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: They were, but I ordered all the JSON objects together. I can add an ordinal to the these objects so its more clearly configurable

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: No worries, @mannyqi ... I do agree - but I think we need to fix that by splitting up in to more categories and having things in a logical order, (and less confusing looking in general)

I prefer scrolling to collapsing / uncollapsing for this sort of thing

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: I think we should just sort in to the same order as specified in that list (rather than having an extra ordinal).

(i.e. use the array index of the item as the ordinal)

matthampton-stock2shop: And re having the JSON objects together at the top - I think we should rather let them be ordered along with the order settings. 

(The fact that they are structured values doesn't make them so special that they should get to be at the top)

justinengle-krs: Okay cool sounds good.

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Sorted

justinengle-krs: Done

justinengle-krs: @matthampton-stock2shop I dont see the options coming back in the template. not sure if it makes sense to hard code these?

image
image

matthampton-stock2shop: Are there any that give options? I feel like there is a way to express that in the metatemplate config, right? Maybe we just haven't done it for these ones?

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: None of them in the trade store config seem to have ? I am not sure which property it would be? values?

Also, I am not sure how you would configure that on the tempalte screen? I dont have access to it, but just based on this screenshot

https://user-images.githubusercontent.com/124143173/246091633-164e53dc-8695-45aa-9744-9d4ab61e41bf.png

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Its based on the type on the meta tempate, if its a boolean -> then toggle, so this is configurable from the template

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: @matthampton-stock2shop I have made this configurable and updated to the best of my knowledge. Let me know if you have any better suggestions

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -3,7 +3,6 @@
...
   import { location } from 'svelte-spa-router';
   import IconFilter from '@/components/Icons/IconFilter.svelte';
   import IconLogistic from '@/components/Icons/IconLogistic.svelte';
```
matthampton-stock2shop, line 8: Is it possible to get the placeholder text to be a placeholder? (rather than a default value)

image

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -16,9 +15,28 @@
   import IconNotification from '@/components/Icons/IconNotification.svelte';
   import IconDomain from '@/components/Icons/IconDomain.svelte';
```
matthampton-stock2shop, line 16: You can add checkout fields but there's no way to remove one:

image

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -16,9 +15,28 @@
...
   import IconTerms from '@/components/Icons/IconTerms.svelte';
-  import channelMetaStore from '@/lib/channelStore';
+  import TextInput from '@/components/input/TextInput.svelte';
```
matthampton-stock2shop, line 18: For the options on the drop down type of checkout field:

image

Can we make the have "Label" instead of "Key"?

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -16,9 +15,28 @@
...
+  import CheckoutFields from './ConfigItems/Forms/CheckoutFields.svelte';
+  import PaymentMethods from './ConfigItems/Forms/PaymentMethods.svelte';
```
matthampton-stock2shop, line 20: Can we move the "order settings" up to just after 'checkout':

image

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: I've logged an issue for us to follow up on this: #196

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: logged issue to follow up: #196

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: Seems like I can still do this:

image

(It's not as confusing now - cos previously the expand/contract actions would apply to all of the repeats - but this still applies to delete at the moment, and it doesn't really make sense to duplicate - so I don't think we should allow it)

justinengle-krs: @matthampton-stock2shop I think I got this sorted.  Let me know?

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
justinengle-krs: Done

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -16,43 +15,244 @@
...
+  import Loader from '../Loader/Loader.svelte';
+  import KeyValueForm from './ConfigItems/Forms/KeyValueForm.svelte';
```
matthampton-stock2shop, line 22: One more little thing, Justin...

image

Could you move the "Add Shipping Method" button to the bottom? It feels odd to click it up there and then the new one gets added down at the bottom where you can't see it

matthampton-stock2shop (approved):

PR #199, opened 2023-07-21
Customer - should only be able to add to "Trade" channels

justinengle-krs:
Only allow trade channels


mannyqi:
@justinengle-krs I noticed the warehouse dropdown is not showing a description, but `[object object]` instead. But there is a todo in there... Just mentioning it in case a solution was already discussed. Let me know if I should ignore for now and I'll approve.

image

mannyqi:
Ignore my previous comment. I see it is fixed in [here](https://github.com/stock2shop/console_v2/pull/198)

mannyqi (approved):

Issue #224, opened 2023-08-16
Settings / Product Templates Issues


justinengle-krs:
Settings / Product Templates 

Scroll Overflow Issue
![image](https://github.com/stock2shop/console_v2/assets/142212593/3a862f1e-9f59-45b5-87a7-1ab44017dd73)

Attribute count defaults to 5 regardless of actual count
![image](https://github.com/stock2shop/console_v2/assets/142212593/0c600cf5-6221-4fb9-b7ec-35d7bd513a54)

_Originally posted by @tauriqdolleyKRS in https://github.com/stock2shop/console_v2/issues/216#issuecomment-1678857145_

Issue #208, opened 2023-08-14
When trying to "manage a template", all of the types are greyed out for existing template attributes


matthampton-stock2shop:
When trying to manage a template, all of the types are greyed out for existing template attributes:

![image](https://github.com/stock2shop/console_v2/assets/124143173/3e1a12fd-486f-4237-a6bb-2e0cc5615503)

If I add a new one I can choose the type.

_Originally posted by @matthampton-stock2shop in https://github.com/stock2shop/console_v2/pull/197#discussion_r1285755878_

justinengle-krs:
As discussed with @matthampton-stock2shop 
The API does not allow editing of these values, so this is intended behaviour, but lets make sure the dropdown displays the value even if its disabled

PR #225, opened 2023-08-16
#224 + #208 - Product Template issues

justinengle-krs:
fix loading state
fix some column spacing
fix attribute count

**Can't replicate having the disabled dropdowns not display the selected value**


casperlotter:
@justinengle-krs I updated this PR.

Screenshot 2023-08-21 102201

src/models/products/product-attribute.ts
```diff
@@ -1,7 +1,15 @@
...
 import type SystemMetaTemplateItem from '@/models/product-templates/system-meta-template-item';
 
-export const ATTRIBUTE_TYPES = { text: 'string', number: 'number', select: 'dropdown', toggle: 'yes / no', json: 'json' };
+export const ATTRIBUTE_TYPES = {
+  text: { label: 'String', value: 'string' },
+  number: {
+    label: 'Number', value: 'number',
+  },
+  select: { label: 'Dropdown', value: 'dropdown' },
+  toggle: { label: 'Yes / No', value: 'yes / no' },
```
justinengle-krs, line 10: so this means that it will save as yes/no to the db? Where the old console saves this as 'boolean'. I see thats catered for in the constructor above. But why the change?

src/components/Settings/ProductTemplate.svelte
```diff
@@ -84,72 +96,79 @@
...
-      type: results.every((p) => !p.error) ? 'success' :
-              results.every((p) => !p.result) ? 'error' : 'info' ,
-      message: results.every((p) => !p.error) ? `Successfully ${action} template.` :
-                 results.every((p) => !p.result) ? results.filter((p) => p.error)[0].error :
-                `Successfully ${action} however some attribute(s) had an error. ${results.filter((p) => p.error).map((m) => m.error).join(' , ')}`,
+      type: results.every((p) => !p.error) ? 'success' : results.every((p) => !p.result) ? 'error' : 'info',
+      message: results.every((p) => !p.error)
+        ? `Successfully ${action} template.`
+        : results.every((p) => !p.result)
+        ? results.filter((p) => p.error)[0].error
```
tauriqdolleyKRS, line 148: Can we guarantee there'll be an zero index error here? Quite a comprehensive set of ternary's..

src/components/Settings/ProductTemplates.svelte
tauriqdolleyKRS, line 1: image

Overflow scroll issue still seems to be present. Attribute count working correctly.

tauriqdolleyKRS (changes requested):
Left some comments. 

Saw comments related to dropdown defaulting. That seems to be working as per those comms.

src/models/products/product-attribute.ts
casperlotter: @justinengle-krs @matthampton-stock2shop I updated it to set the value to boolean instead.

tauriqdolleyKRS (approved):

mannyqi:
Looks like the deletion of the template is missing a confirm.

src/components/Settings/ProductTemplates.svelte
```diff
@@ -79,16 +84,25 @@
...
+            }}
+            size="small"
+            variant="gray"
+            Prefix={IconEdit}
+            iconSize="extra-small"
+            class="mr-4 hover:!bg-neutral-100 !text-neutral-700"
+          />
           <Btn
             onClick={async () => {
               await deleteProductTemplate(row.name);
```
mannyqi, line 108: Looks like the deletion of the template is missing a confirm. i.e. `confirm('Are you sure you want to delete this template?')`

mannyqi (changes requested):

src/components/Settings/ProductTemplates.svelte
casperlotter: @mannyqi  This is called in the deleteProductTemplate function now

mannyqi:
@justinengle-krs when I click the "New Template" button, the modal just pops up with the loader and it just stays that way. There are no errors in the console either. Can you check if the same happens your side?

src/components/Settings/ProductTemplate.svelte
```diff
@@ -165,7 +174,7 @@
     queryClient.invalidateQueries(['meta_template-get']);
   }
 
-  $: isLoading = $templateDetails.isLoading || $templateDetails.isFetching || loading;
+  $: isLoading = (!!productTemplateToEdit && ($templateDetails.isLoading || $templateDetails.isFetching)) || loading;
```
TomRadford, line 177: isLoading (which is just a shorthand for status === 'loading') indicates that "you have no data yet" in Query v4
(https://tkdodo.eu/blog/status-checks-in-react-query) 
Thus, if disabled, isLoading will be true. 

Knowing this, I believe simplifying this to
` $: isLoading = (!!productTemplateToEdit && $templateDetails.isLoading) || loading;`
will do the trick 🎉

src/components/Settings/ProductTemplates.svelte
```diff
@@ -25,7 +25,7 @@
 
   const queryClient = useQueryClient();
 
-  const productTemplateItems = createQuery(['meta_template-get'], () => productTemplateRepo.getAll());
+  $: productTemplateItems = createQuery(['meta_template-get'], () => productTemplateRepo.getAll());
```
TomRadford, line 28: not sure if this one needs to be reactive since it doesn't have any dependencies. (ie: it's just got the queryFn)

src/components/Settings/ProductTemplate.svelte
```diff
@@ -31,37 +32,33 @@
...
+          })
+        );
+    },
+    onError(err) {
+      displayInlineNotification = {
+        type: 'error',
+        message: 'There was an error loading the product template',
+      };
+    },
+    enabled: !!productTemplateToEdit && !templateItemsVC,
```
TomRadford, line 60: You can probably revert to `enabled: !!productTemplateToEdit ` if you go with my suggestion below 👌

TomRadford (changes requested):

mannyqi:
@justinengle-krs let me know when I can review again

justinengle-krs:
@mannyqi @TomRadford Please review and test

mannyqi (approved):

mannyqi (approved):

src/components/Modal/CenterModal.svelte
```diff
@@ -12,7 +12,7 @@
     {#if isLoading}
       <Loader />
     {:else}
-      <div class="flex items-center px-6 py-4 border-b border-neutral-200 bg-white sticky top-0">
+      <div class="flex items-center px-6 py-4 border-b border-neutral-200 bg-white sticky top-0 z-10">
```
TomRadford, line 15: @justinengle-krs nit: was curious why we have both this component, as right Modal and then a generic `Modal` component that supports center, left and right alignment. I'd like to refactor all of these into `Modal` at some point but just want to double-check check there isn't another reason?

TomRadford (approved):
Looks good, just left a comment RE: modals :)

src/components/Modal/CenterModal.svelte
justinengle-krs: @TomRadford There was no important reason. That original modal was done by the previous devs - when I started on here, I think it was just easier to create that side modal from scratch to meet my needs at the time, and then subsequently someone else made the center one.
So I think no harm in refactoring them into 1 :)

PR #233, opened 2023-08-17
Swop range values if to is smaller than from

casperlotter:


tauriqdolleyKRS:
Swop is a nice QoL change but there's still a filtering bug. 

image

tauriqdolleyKRS (approved):

PR #252, opened 2023-08-24
Catch unhandled exceptions and display it to the user

casperlotter:
Screenshot 2023-08-24 154813


matthampton-stock2shop:
I'm not sure I like this approach - is there not a Svelte way to do this?

matthampton-stock2shop:
> I'm not sure I like this approach - is there not a Svelte way to do this?

@TomRadford @justinengle-krs

TomRadford:
@matthampton-stock2shop it depends on what context the error are coming from, Svelte Query Client has a global onError config option for the QueryCache config we could use for any request-related errors:
`export const queryClient = new QueryClient({
  defaultOptions: { queries: { retry: 1 } },
  queryCache: new QueryCache({'onError': () => alert('woops!')})
});
`
as per TKDodo's recommendation here: https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose#a-bad-api
scroll down to where he talks about how setting global cache-level callbacks is a great solution. 

Outside of the query context where we need to catch other errors (ie: accessing undefined properties) that causes the application to throw - a pattern like this looks pretty nice: https://khromov.se/error-boundaries-in-svelte/ which uses this old package https://github.com/crownframework/svelte-error-boundary

Sadly, native error boundaries seem to be something missing from svelte 3/4 and I it's something that may be addressed in V5 (the community seems to want them and there's a heated discussion about it here https://github.com/sveltejs/rfcs/pull/46#issuecomment-1536616132) 
My money is on some sort of error handling API to be introduced when v5 comes out

PR #259, opened 2023-08-28
defaultQuery wrapper for useQuery

casperlotter:
Wrap useQuery in s2s default wrapper where we can specify default options


src/lib/default-query.ts
```diff
@@ -0,0 +1,61 @@
...
+  UseQueryStoreResult,
+} from '@sveltestack/svelte-query';
+
+export default function defaultQuery<TQueryFnData = unknown, TError = unknown, TData = TQueryFnData, TQueryKey extends QueryKey = QueryKey>(
+  arg1: TQueryKey | UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+  arg2?: QueryFunction<TQueryFnData, TQueryKey> | UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+  arg3?: UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+): UseQueryStoreResult<TQueryFnData, TError, TData, TQueryKey> {
+  const defaultConfig = {
+    retry: 1,
```
TomRadford, line 15: Could one not rather set up shared options when instantiating the QueryClient, for example with:
`export const queryClient = new QueryClient({ defaultOptions: { queries: { refetchOnWindowFocus: import.meta.env.DEV ? false : 'always', retry: 1 } } }); `

All instances of queries in child components of the `QueryClientProvider` with this `queryClient` as the `client` prop will make use of this queryClient and thus share its config 🎉

You can see an example of this in my PR here: https://github.com/stock2shop/console_v2/pull/260/files#r1307499575

TomRadford:
Closing since https://github.com/stock2shop/console_v2/pull/260/commits/87927d5c35acfd6b2b61bb9bce403a5df6ff60ff will resolve this issue

TomRadford:
Closing since https://github.com/stock2shop/console_v2/pull/260/commits/87927d5c35acfd6b2b61bb9bce403a5df6ff60ff will resolve this issue

Issue #212, opened 2023-08-14
Segments - Show Data Source


justinengle-krs:
![image](https://github.com/stock2shop/console_v2/assets/125956275/0e34fffb-878c-4250-99ad-4eefa1c12145)

TomRadford:
@matthampton-stock2shop Keen to chat about the data source and functionality for this drop down.

TomRadford:
Functionality notes from chat with Matt:

- Data source mapped from `segment.owner`
- This page is most valuable from a viewing perspective for Source segments
- Data sources should only be mutable if they're set on S2S (owner === `system`)

PR #262, opened 2023-08-29
#212 & #211: Segments - Show Data Source & update/delete functionality

TomRadford:
Resolves: #212 #211 
Design:
image

Implementation:
image

Segment table features:
- data source display
- 'add' functionality
- 'delete' functionality

Additional changes in this PR:
- add segment row header and delete
- add VS Code tailwind intellisense for cn util
- add additonal cn doc
- use existing data instead of invalidating state


src/components/Segments/SegmentsList.svelte
```diff
@@ -16,28 +16,54 @@
...
+    openModal(context, true, segment);
+  }
+
+  function openModal(context: SegmentType | MouseEvent, edit?: boolean, segment?: SystemSegment) {
     if (isSegmentType(context)) {
       modalContext = context;
     }
     inlineNotification = null;
-    addSegmentModel = new SystemSegment({ user_id: userId, owner: 'system', type: 'products' });
+    mutateSegmentModel = edit ? segment : new SystemSegment({ user_id: userId, owner: 'system', type: 'products' });
```
TomRadford, line 37: Allows us to share this this model for both update and delete operations,

src/components/Segments/SegmentsList.svelte
```diff
@@ -16,28 +16,54 @@
...
+      editModal = true;
+    } else {
+      editModal = false;
+    }
     modalOpen = true;
   }
 
   async function addSegment() {
-    const segmentsPut = [...segments.filter((s) => s.type == addSegmentModel.type), addSegmentModel];
+    const segmentsPut = [...segments.filter((s) => s.type === mutateSegmentModel.type), mutateSegmentModel];
```
TomRadford, line 47: ensure quality check is strict ie: `===`

src/components/Segments/SegmentsList.svelte
```diff
@@ -16,28 +16,54 @@
...
     const result = await segmentsRepo.put(new SystemSegmentDataWrapper({ system_segments: segmentsPut }));
     inlineNotification = { type: result.result ? 'success' : 'error', message: result.result ? 'Successfully added segment.' : result.error };
-    queryClient.invalidateQueries(['segments-get', userId]);
+    if (!!result.result && typeof (result.result as SystemSegment[])?.length === 'number') {
+      queryClient.setQueryData(
+        ['segments-get', userId],
+        [...segments.filter((s) => s.type !== mutateSegmentModel.type), ...(result.result as SystemSegment[])]
```
TomRadford, line 53: Added this here instead of invalidating the query as a means of saving on an additional `get` (during invalidation) since all the data we need already exists and we simply need to mutate the state accordingly.

src/components/Segments/SegmentsList.svelte
```diff
@@ -16,28 +16,54 @@
...
+    }
+    modalOpen = false;
+  }
+
+  async function editSegment() {
+    const segmentsEdit = [...segments.filter((s) => s.type === mutateSegmentModel.type || s.key === mutateSegmentModel.key), mutateSegmentModel];
+    const result = await segmentsRepo.put(new SystemSegmentDataWrapper({ system_segments: segmentsEdit }));
+    inlineNotification = { type: !!result.result ? 'success' : 'error', message: !!result.result ? 'Successfully added segment.' : result.error };
+    if (!!result.result) {
+      queryClient.setQueryData(['segments-get', userId], segments);
```
TomRadford, line 64: Added this here instead of invalidating the query as a means of saving on an additional get (during invalidation) since the mutated segments array will already exist as a result of the mutation of `segments` as per `bind:segment={mutateSegmentModel}` (mutateSegmentModel is a ref to segment when editing) in `AddSegmentModal`

src/components/Segments/SegmentsList.svelte
```diff
@@ -68,10 +101,10 @@
 {:else}
   <div class="flex flex-col gap-8">
     {#if segmentContext === 'users'}
-      <SegmentsType type="customers" on:click={() => openModal('customers')} {segments} />
-      <SegmentsType type="products" on:click={() => openModal('products')} {segments} />
+      <SegmentsType type="customers" on:click={() => openModal('customers')} {segments} {segmentsRepo} {userId} {openEditModal} />
```
TomRadford, line 104: Instead of prop drilling, we could also make use of the [context api](https://learn.svelte.dev/tutorial/context-api), however, this implementation favours simplicity and, thus verbosity, over complexity.

src/components/input/Dropdown.svelte
```diff
@@ -21,17 +21,24 @@
...
     }
   });
-
-  $: {
-    if (onValueChanged) onValueChanged(value);
-  }
 </script>
 
-<div class="flex  relative {className} {sideLabel ? 'flex-row items-center justify-between' : 'flex-col'}">
+<div class={cn(`flex relative`, sideLabel ? 'flex-row items-center justify-between' : 'flex-col', className)}>
```
TomRadford, line 26: refactoring to use our new `cn` util

src/repos/segments-repo.ts
```diff
@@ -18,14 +18,32 @@ export default class SegmentsRepo {
...
+  async put(segments: SystemSegmentDataWrapper): Promise<ApiResult<boolean | SystemSegment[]>> {
     try {
       const opts: APIRequest<SystemSegmentDataWrapper> = {
         path: this.basePath,
         body: segments,
       };
+      const res = await put<SystemSegmentDataWrapper>(opts);
+      return {
+        error: '',
+        result: res.system_segments,
```
TomRadford, line 30: since we're getting data returned from this endpoint, I thought it would be valuable to return it here in order to mutate state (as noted above where I've used this changed to get rid of the query invalidation)

src/components/Segments/SegmentsList.svelte
casperlotter: @TomRadford @matthampton-stock2shop @justinengle-krs 

My thoughts on this:

The GET request might ensure the API call for this resource is cached on the server.
The additional complexity introduced here in the code. For example, if similar code is run in the context of a ObjectVersionControl it might cause unintended side effects. So I would be careful to not make this a pattern we apply in general.

TomRadford: Thanks, Casper, @matthampton-stock2shop, I'm keen to get your opinion on this.
We could possibly invalidate all queries when ObjectVersionControl changes by means of changes to the `queryClient` config.

mannyqi:
@TomRadford 

I picked up a small bug where the name of the customer becomes undefined:
image


**Steps to reproduce:**
1. Add a segment
2. Click/move away from the page by clicking on another screen or browser tab
3. Click back on the segment page

Note that the customer name changes to "undefined undefined" after the content refreshes.

**Also, a couple of questions:**
1. Should the deletion of segments not have a confirmation?
2. Is the "chevron-down" icon needed for the data source? From the looks of it, we can no longer change the data source manually?
image

TomRadford:
> @TomRadford
> 
> I picked up a small bug where the name of the customer becomes undefined: image
> 
> **Steps to reproduce:**
> 
> 1. Add a segment
> 2. Click/move away from the page by clicking on another screen or browser tab
> 3. Click back on the segment page
> 
> Note that the customer name changes to "undefined undefined" after the content refreshes.
> 
> **Also, a couple of questions:**
> 
> 1. Should the deletion of segments not have a confirmation?
> 2. Is the "chevron-down" icon needed for the data source? From the looks of it, we can no longer change the data source manually?
> 
> image

@mannyqi should be resolved now.  That undefined issue was resolved by syncing with master which included some useful query improvements.

src/stories/components/80-Segments/10-SegmentRow.stories.svelte
```diff
@@ -5,17 +5,21 @@
   import SegmentRow from '@/components/Segments/SegmentRow.svelte';
   import { configLocal } from '@/env/config.local';
   import type SystemSegmentDataWrapper from '@/models/segments/system-segment-data-wrapper';
+  import { QueryClient, QueryClientProvider } from '@sveltestack/svelte-query';
```
mannyqi, line 8: Storybook build failed due to svelte-query not being found.

image

mannyqi (changes requested):
Other than this one small fix, everything else seems to be working.

mannyqi (approved):

src/components/form/Btn.svelte
```diff
@@ -4,14 +4,16 @@
...
+  import { cn } from '@/utils/classname';
 
   type SvelteComponentConstructor = new (...args: any[]) => SvelteComponent;
 
   export let Prefix: SvelteComponentConstructor | IconUnion | undefined = null;
   export let Suffix: SvelteComponentConstructor | IconUnion | undefined = null;
 
   export let disabled = false;
-  export let size: 'small' | 'medium' | 'large' = 'medium';
+  export let size: Size = 'medium';
```
matthampton-stock2shop, line 16: I'm not 100% sure why, but it seems like maybe these changes have affected how the buttons font looks (it's become big)... Could obviously be somewhere else, too... not sure:

image

TomRadford: 100% this PR broke button sizing

PR #268, opened 2023-08-31
Small copy and code cleanup

casperlotter:
Screenshot 2023-08-31 131847


src/routes/Products.svelte
```diff
@@ -196,7 +196,7 @@
 </script>
 
 <div>
-  <ImportModal context="product" bind:open={showImportProductModal} {uploadComplete} />
+  <ImportModal context="Product" bind:open={showImportProductModal} {uploadComplete} />
```
TomRadford, line 199: Instead of using this, it might be better to use our capitalise util where capital is needed.

TomRadford, line 199: Instead of using this, it might be better to use our capitalise util where capital is needed.

src/components/Import/ImportModal.svelte
```diff
@@ -12,7 +12,7 @@
   import { getFileSizeString } from '@/lib/file-utils';
 
   export let open = true;
-  export let context: 'product' | 'customer';
```
TomRadford, line 15: Any reason for loosening the rigidity of this type by removing the literal union? 
Looking at the designs, I think it might be easier to keep this as is for customer/product imports

TomRadford (changes requested):

src/components/Import/ImportModal.svelte
```diff
@@ -67,7 +67,7 @@
   <div slot="content" class="p-6 text-neutral-700 text-regular">
     <p class="mb-4">
       You can bulk add and update {pluralizeWord(context)} by importing a csv file. You can download a
-      <a href="#" class="font-bold text-brand-notification"> sample file</a>
+      <span class="font-bold text-brand-notification cursor-pointer">sample file</span>
```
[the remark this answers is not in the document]
TomRadford: resolved" Matt's answer = not yet.

TomRadford (approved):
made changes and approved

Issue #265, opened 2023-08-30
Re-enable refetchOnWindowFocus on dev (or disable it totally)


matthampton-stock2shop:
As discussed we are going to re-enable refetchOnWindowFocus (or disable it totally - so that it is not different between dev and prod)

_Originally posted by @matthampton-stock2shop in https://github.com/stock2shop/console_v2/pull/260#discussion_r1309850379_

TomRadford:
This ticket also involves replacing ui loading state indicators to use `isLoading` instead of `isFetching` to prevent UI flickers when refocusing the window.

TomRadford:
As per chat with Chris, going to look into skeleton loaders as part of this ticket.

TomRadford:
After some investigation, I believe using skeletons might not make sense from a user perspective on the table components since the current implementation of lowering the opacity of the table when running a new query works well to indicate new results are being loaded:
![Image](https://github.com/stock2shop/console_v2/assets/7515754/9499e477-dafc-4db3-b757-60e484f84368)

PR #271, opened 2023-09-01
#265: loading state changes & reactive declaration change

TomRadford:
This fixes:
- flashing to loading ui on settings/users/[id]/segments and in a couple of components where indicating refetching is not valuable. 
- reactive declarations to enable reactive state on query


src/components/Segments/SegmentsList.svelte
```diff
@@ -61,7 +61,7 @@
   </div>
 {/if}
 
-{#if isFetching}
+{#if isLoading}
```
TomRadford, line 64: example of where I switched out to fetching to prevent unnecessary ui flashes - the user will see the updated data 
(if anything changes on the server) once it's been fetched.

src/components/Customer/CustomerDetail.svelte
```diff
@@ -51,21 +51,22 @@
     },
   });
 
-  const customerData = createQuery(['customer-get', customerId], () => customersRepo.get(customerId), {
+  $: customerData = createQuery(['customer-get', customerId], () => customersRepo.get(customerId), {
     onSuccess(data) {
       const systemCustomerUser = new SystemCustomer(data);
       if (!customerVersionControlled) customerVersionControlled = new ObjectVersionControl<SystemCustomer>(systemCustomerUser);
```
TomRadford, line 57: @justinengle-krs changing these queries to reactive declarations allows them to be reactive to any dependant variables, in this case it seems to be the `*VersionControlled` dependency that was causing the issue if these queries where not flagged as reactive.

src/components/Configurators/ColumnConfigurator/ColumnConfigurator.svelte
```diff
@@ -83,17 +83,14 @@
...
-  const result = createQuery([`meta-fields-${index}`, debouncedValue], () =>
+  $: result = createQuery([`meta-fields-${index}`, debouncedValue], () =>
     doPostElasticSearch($request.path, body, {
       exclude_products: true,
     })
   );
   const updateSearch = () => {
     debouncedValue = filterValue;
     body = metaFieldsSearchRequest(debouncedValue, $request);
-    result.updateOptions({
```
TomRadford, line 94: @justinengle-krs anywhere where we were using this update options attribute has now been removed and thus the query should be declared with a reactive declaration, ie:  `$:` instead of `const` as per https://tanstack.com/query/latest/docs/svelte/reactivity

matthampton-stock2shop (approved):

Issue #244, opened 2023-08-21
Info display does not transition back to a gray button once you have saved:


casperlotter:
Info display does not transition back to a gray button once you have saved:

![Screenshot 2023-08-14 162729](https://github.com/stock2shop/console_v2/assets/129734011/10e82067-30b7-4d38-a0bd-fb1735f9b121)

_Originally posted by @casperlotter in https://github.com/stock2shop/console_v2/issues/216#issuecomment-1677426287_

TomRadford:
Got a bugfix ready just waiting for @justinengle-krs's feedback on https://github.com/stock2shop/console_v2/commit/1ec1c8d69c99d45f122c377dbe4f109d305930c2#r126690552 and https://github.com/stock2shop/console_v2/commit/f31e21231b0c522880586afbc65d90dc0edb0994#r126696527

PR #276, opened 2023-09-07
#244: Save changes state disabled state fixed 

TomRadford:
- Renamed schema state to include `schema` to prevent confusion
- Changed `changedMade` into a normal state declaration with `let` instead of `$:`
- mutate changes made in the same reactive code block where channelMetaValuesArray gets mutated (line 313) this is in order to prevent svelte batching microtasks and not updating the changesMade before the rest of the UI updates when updated in it's own $: block  
- cleanup and document the VersionControlled class


src/components/Connectors/ConfigItems/ProductFilters.svelte
```diff
@@ -111,7 +111,7 @@
 
   $: availableFields = $availableFieldsData.data || [];
 
-  $: isLoading = loading || $availableFieldsData.isLoading || $rulesData.isLoading || $availableFieldsData.isFetching || $rulesData.isFetching;
+  $: isLoading = loading || $availableFieldsData.isLoading || $rulesData.isLoading;
```
TomRadford, line 114: isFetching is not needed here as it causes loader flash when stale query data is being revalidated

src/models/version-controlled.ts
```diff
@@ -1,6 +1,17 @@
 import stableStringify from 'json-stable-stringify';
 import omit from 'lodash.omit';
 
+/**
+ * Svelte Store with version control utilities.
+ * Useful for validating form data against the server
+ * to check for changes prevent overwriting new server data
+ *
+ * @example
+ * const serverData = new ObjectVersionControl<ServerData>(initialData)
```
TomRadford, line 10: popped this here so that I remember this is a custom store implementation and should be subscribed to 😝

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -464,6 +446,7 @@
           bind:changesMade={prodFiltersChangesMade}
         />
       {:else if !loading && channelMetaValuesArray?.length}
+        <!-- TODO: Remove -->
```
TomRadford, line 449: As per https://github.com/stock2shop/console_v2/commit/f31e21231b0c522880586afbc65d90dc0edb0994#r126696527 I want to clean this up when there's time

src/models/version-controlled.ts
```diff
@@ -78,7 +80,7 @@ export default class ObjectVersionControl<T> {
       // the user has not made changes and the information has updated on the server
       // so update the working copy and the base server version
       this.baseServerVersion = newServerVersion;
-      this.workingCopy = this.parseArray(newServerVersionJson);
+      this.workingCopy = JSON.parse(this.stringify(newServerVersionJson)) as T;
```
matthampton-stock2shop, line 83: This looks wrong to me - that it is stringifying the newServerVersionJson - it should just be parsing it...

I'm not sure how this has been working TBH cos it is going to result in having string instead of an object for the working copy.. Or am I mistaken? 

Maybe no-one has seen the issue because it requires you to make a change on the server while the console is open to hit this part of the code.

i.e. open the console in two tabs to the same screen

And make a change on one, then go to the other and see it break

Would you mind checking for me?

matthampton-stock2shop (changes requested):

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -305,25 +307,12 @@
...
-        const parsedValue = JSON.parse(value);
-        if (typeof parsedValue === 'object') {
-          return JSON.stringify(parsedValue);
-        }
-      } catch (error) {
-        // Invalid JSON string, do nothing and continue
-      }
-    }
-    return value;
+    changesMade = channelMetaValuesArray.some((c) => c.changesMade) || prodFiltersChangesMade || prodFiltersChangesMade;
```
matthampton-stock2shop, line 313: I've also still got a bit of funny behaviour with the save changes button (sorry):

https://github.com/stock2shop/console_v2/assets/124143173/aadd6877-8f27-4418-8a3e-98ba4711c7e2

src/models/version-controlled.ts
TomRadford: Can confirm, this should have been `this.workingCopy = JSON.parse(newServerVersionJson) as T;`

TomRadford: Seems that after checking this seem to be getting state on channelMetaValuesArray behaviour that isn't reactive instantly reactive.
https://github.com/stock2shop/console_v2/assets/7515754/257ac4ef-412f-4bed-b503-ca15a8ae0422

@matthampton-stock2shop perhaps I should try action that refactor today (into a single VersionControlled store) to reduce complexity and allow for (hopefully) more predictive reactive behaviour.

TomRadford:
Chatted with Matt and we're planning to tackle a refactor on Monday using a singular VersionControlled store as per https://github.com/stock2shop/console_v2/pull/276#discussion_r1319507812

src/components/Connectors/ChannelMaster.svelte
matthampton-stock2shop: We're going to sort this out separately

matthampton-stock2shop (approved):

PR #279, opened 2023-09-13
Fix 'save changes' button being dodgy on the channel settings pages (and simplify code)

matthampton-stock2shop:
This fixes the save changes button toggling active and inactive in a semi-random fashion on the channel settings pages by doing a fairly large refactor to have a single ObjectVersionControl instance that contains a record (a map from the meta setting key to its value).

We also fixed a bug in the ObjectVersionControl that stopped it from accurately reporting isStale and changesMade when accessed as a $store.

After that we've done a re-structure of the code in order to have more straightforward svelte components for each settings screen - rather than generating them from the meta template data directly (with lots of hooks to customise).

This makes it easier to build more complex things (e.g. that use / set more than one setting like the Domain settings screen. (That is now far simpler).

The ProductFilters screen is also now not a 'special' case (where the saving / changesMade code was happening in the master screen behind special if blocks). Instead that code is now where it belongs.

The refactor also changes the routing to be more straightforward, and reduces boilerplate copy and paste in lots of places.


src/components/Connectors/ChannelMaster.svelte
```diff
@@ -140,289 +118,149 @@
       ],
     },
     { navigation: 'domain', label: 'Domain', IconComponent: IconDomain, fields: ['domain_alias', 'login_redirect'] },
-    { navigation: 'terms-conditions', label: 'Terms & Conditions', IconComponent: IconTerms, fields: ['email', 'terms'] },
+    { navigation: 'terms-conditions', label: 'Terms & Conditions', IconComponent: IconTerms, fields: ['terms'] },
```
matthampton-stock2shop, line 121: I would would work through this NavItems list one by one and create a setting screen component for each one, and then remove the fields from this list so that in the end we only have the label, icon etc. in here

src/components/Connectors/ChannelMetaSetting.svelte
```diff
@@ -0,0 +1,117 @@
...
+    import { getContext } from 'svelte';
+
+    export let settingKey: string;
+
+    const settings = getContext<ObjectVersionControl<Record<string, string>>>('settings');
+    $: metaTemplate = getContext<Record<string, SystemMetaTemplateItem>>('metaTemplateLookup')[settingKey];
+
+    const componentDictionary = {
+
+      // checkout_fields: {
```
matthampton-stock2shop, line 18: As you go, come and look here and bring back the special custom components in to the setting screen pages where they belong

src/components/Connectors/ChannelMetaSetting.svelte
```diff
@@ -0,0 +1,117 @@
...
+
+</script>
+
+<div class="w-full h-fit mb-4 block">
+    <p class="text-smaller leading-4 text-brand-action uppercase font-bold tracking-widest my-6">
+        {formatKey(settingKey)}
+    </p>
+
+    {#if componentDictionary[settingKey]}
+        <svelte:component
```
matthampton-stock2shop, line 95: So once we're done, this section can be gone

src/components/Connectors/ConfigItems/DomainAlias.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+  import Notification from '@/components/Product/Notification.svelte';
+  import TextInput from '@/components/input/TextInput.svelte';
+  import ObjectVersionControl from '@/models/version-controlled';
+  import {getContext} from "svelte";
+
+  const settings = getContext<ObjectVersionControl<Record<string, string>>>('settings');
```
matthampton-stock2shop, line 7: Then we will want to have working storybook stories for these components. They should be not too hard to make - cos you can set these contexts in the story with mock data.

('settings' and 'metaTemplateLookup')

src/components/Connectors/ConfigItems/DomainAlias.svelte
```diff
@@ -0,0 +1,18 @@
...
+
+  $: {
+    // These must be kept in sync for some reason
```
justinengle-krs, line 10: This spec was provided by Chris. I can't recall the exact reasoning

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
```diff
@@ -24,57 +29,63 @@
...
     // Insert the property at the new index
     properties.splice(newIndex, 0, propertyToMove);
-    propertyValues.splice(newIndex, 0, productFilters[propertyToMove]);
+    propertyValues.splice(newIndex, 0, $settings.workingCopy.search_facets[propertyToMove]);
 
     // Create the updated object with reordered properties
     const updatedProductFilters = properties.reduce((obj, property, index) => {
-      obj[property] = propertyValues[index];
+      // add the index here so that VersionControlObject can detect changes
+      obj[property] = { ...propertyValues[index], index };
```
TomRadford, line 49: @matthampton-stock2shop had to add the index here so that our vc diff doesn't pick up a re-order as having no changes. Just confirming API will be happy with this or shall I figure out a way to filter out this index attribute after the diff in VC?

TomRadford:
@matthampton-stock2shop looks like everything is complete on this one. Just keen for feedback on https://github.com/stock2shop/console_v2/pull/279/files#r1324714458 ... then after a once-over we can (hopefully) merge 🥳

src/routes/Channel.svelte
```diff
@@ -3,7 +3,7 @@
   import { push } from 'svelte-spa-router';
   import ChannelMaster from '@/components/Connectors/ChannelMaster.svelte';
 
-  export let params: { id: number, settingGroup: string };
+  export let params: { id: number; settingGroup: string };
```
matthampton-stock2shop, line 6: What does this change do? Is it supposed to be a semi-colon in this context?

src/stories/components/30-Connectors/ChannelSettings/ChannelSettings.stories.svelte
```diff
@@ -0,0 +1,114 @@
...
+
+  const channelsRepo = new MockChannelsRepo();
+  const queryClient = new QueryClient();
+
+  const settings: ObjectVersionControl<Record<string, string>> = new ObjectVersionControl<Record<string, string>>({});
+
+  let metaTemplateLookup: SystemChannelMetaValues;
+
+  // Copied from channel master for this story group to work
+  function tryParseItLikeAnObject(value: string) {
```
matthampton-stock2shop, line 31: I think we should maybe rather put this in to a utils package or something?

src/stories/components/30-Connectors/ChannelSettings/ChannelSettings.stories.svelte
```diff
@@ -0,0 +1,114 @@
...
+      if (typeof v === 'object') {
+        // This also includes arrays just cos of how JS likes to be weird
+        return v;
+      }
+    } catch (e) {}
+    return value;
+  }
+
+  // Copied from channel master for this story group to work
+  function getSettings(channelMetaList: SystemChannelMeta[]) {
```
matthampton-stock2shop, line 43: And / or this maybe?

src/stories/components/30-Connectors/ChannelSettings/10-ProductFilters.stories.svelte
```diff
@@ -0,0 +1,42 @@
...
+  const productsRepo = new MockProductsRepo();
+
+  const rulesRepo = new MockRulesRepo();
+  const queryClient = new QueryClient();
+
+  const settings: ObjectVersionControl<Record<string, string>> = new ObjectVersionControl<Record<string, string>>({});
+
+  setContext('settings', settings);
+
+  // let metaTemplateLookup: SystemChannelMetaValues;
```
matthampton-stock2shop, line 25: Maybe remove the commented out bit?

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
```diff
@@ -85,19 +84,19 @@
 
   function save(filter, originalLabel) {
     if (!originalLabel) {
-      productFilters[editingFilter.label] = removeLabelProp(filter);
+      $settings.workingCopy.search_facets[editingFilter.label] = removeLabelProp(filter);
```
matthampton-stock2shop, line 87: I'm getting a traceback here adding a new one: 

```
SearchFacets.svelte:87 Uncaught TypeError: Cannot set properties of undefined (setting 'Category')
    at save (SearchFacets.svelte:87:1)
    at HTMLButtonElement.func (SearchFacetsFilterAddEdit.svelte:20:57)
    at HTMLButtonElement.<anonymous> (Btn.svelte:83:20)
```

src/components/Connectors/ConfigItems/Forms/ShippingMethodEditModal.svelte
```diff
@@ -11,7 +10,7 @@
   import IconLogistic from '@/components/Icons/IconLogistic.svelte';
 
   export let open = false;
-  export let save: (methodToEdit: any) => void;
+  export let onOK: (methodToEdit: any) => void;
   export let shippingMethod: any = null;
 
   let taxCodes = ['taxed', 'exempt'];
```
matthampton-stock2shop, line 16: I'm getting an error at line 19 trying to remove a taxline:
```
ShippingMethodEditModal.svelte:19 Uncaught TypeError: Cannot read properties of undefined (reading 'filter')
    at removeTaxLine (ShippingMethodEditModal.svelte:19:1)
    at Array.click_handler (ShippingMethodEditModal.svelte:89:31)
    at HTMLDivElement.click_handler (ShippingMethodEditModal.svelte:92:45)
```

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
```diff
@@ -24,57 +22,58 @@
...
+      obj[property] = { ...propertyValues[index], index };
       return obj;
     }, {});
 
-    //this seems to be a weird svelte quirk, we need to give it time to redraw the array
-    productFilters = null;
-    await sleep(5);
-    productFilters = JSON.parse(JSON.stringify(updatedProductFilters));
+    $settings.workingCopy.search_facets = null;
+    await tick();
```
matthampton-stock2shop, line 54: Do we actually need this? Looks a bit fishy to me

src/components/Connectors/ConfigItems/Forms/ShippingMethods.svelte
```diff
@@ -8,100 +8,70 @@
...
-  }
-
-  $: {
-    let saveObj = {
-      ...shippingMethods,
-      ...newlyAddedMethods.reduce((acc, cur) => {
-        acc[cur.title] = cur;
-        return acc;
-      }, {}),
+      [`NEW_SHIPPING_METHOD_${Object.keys($settings.workingCopy.shipping_methods).length + 1}`]: {
```
matthampton-stock2shop, line 28: Do we need to have this funny looking default name?

image

Can it not be blank to start with? Does it have to have a value cos we're using it as an object prop name?

src/components/Connectors/ConfigItems/Forms/ShippingMethods.svelte
```diff
@@ -110,45 +80,26 @@
     <Notification type="info" text="Shipping methods available to your customers." />
   </div>
 
-  {#if shippingMethods}
-    {#each Object.keys(shippingMethods) as methodProp}
+  {#if $settings.workingCopy.shipping_methods}
+    {#each Object.keys($settings.workingCopy.shipping_methods) as methodProp}
       <div class="w-full text-regular cursor-pointer   my-[16px]  flex justify-between items-center font-bold">
         <TextInput className="w-2/5 mr-2" bind:value={methodProp} />
```
matthampton-stock2shop, line 86: Changing this doesn't seem to have an effect (which I guess makes sense given that it is binding to an iteration var).. Hmmm...

matthampton-stock2shop: Would you mind fixing it? (I know it was obviously already broken)

src/routes/Channel.svelte
TomRadford: yes I believe es lint prefers a semicolon for types

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
TomRadford: Hey Matt, not entirely happy with our implementation of sortablejs so I'm currently looking into something a little less fishy :)

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
TomRadford: Can't seem to replicate this one on my side... could be that my search facets dataset is different? However, have refactored this component quite a bit, so it could be resolved has a side effect.

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
TomRadford: Removed sortablejs and refactored to use svelte-dnd-action ... significantly less fishy now IMO 👍 
Although sortablejs can definitely work great with svelte, I found that svelte-dnd-action provides a really elegant out-of-the-box solution using Svelte Actions - thus getting us where we need to be quickly with less hacks 🥳

src/components/Connectors/ConfigItems/Forms/ShippingMethods.svelte
TomRadford: This was actually broken, should be resolved now :) 
New methods will have empty key and the user will only be able to add an additional key one that key have a name (to prevent duplicate keys)

src/components/Connectors/ConfigItems/Forms/ShippingMethods.svelte
TomRadford: Should be resolved now as per my comment above - not sure how I missed this one!  
Seems to be working nicely now 👍

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
```diff
@@ -78,25 +42,32 @@
...
+  function save(filter: any, originalLabel: string) {
+    //ensure order integrity based on label as key
+    if (listItems.some((item) => item.name === originalLabel)) {
+      const items = Object.entries($settings.workingCopy.search_facets);
+      $settings.workingCopy.search_facets = Object.fromEntries(
+        items.map((item) => (item[0] === originalLabel ? [filter.label, removeLabelProp(filter)] : item))
+      );
     } else {
-      delete $settings.workingCopy.search_facets[originalLabel];
       $settings.workingCopy.search_facets[editingFilter.label] = removeLabelProp(filter);
```
matthampton-stock2shop, line 57: This is the line where it failed for me cos 'search_facets' is null/undefined - needs to be init'd to an empty object in this case

src/components/Connectors/ConfigItems/Forms/SearchFacets.svelte
```diff
@@ -78,25 +42,32 @@
...
 
-  $: availableLabels = [...Object.keys($settings.workingCopy.search_facets ?? []), ...newFilters.map((filter) => filter?.label)] ?? [];
-  $: $settings.workingCopy.search_facets = $settings.workingCopy.search_facets;
-  $: keys = Object.keys($settings.workingCopy.search_facets ?? []);
+  $: availableLabels = Object.keys($settings.workingCopy.search_facets ?? []) ?? [];
+
+  // Generate sorted keys for list (dnd will mutate this value)
+  $: listItems = Object.entries($settings.workingCopy.search_facets ?? [])
+    .sort(([_, a], [__, b]) => {
+      return a[1]?.index ?? 0 - b[1]?.index ?? 0;
```
matthampton-stock2shop, line 68: Let's chat about this, but I'm thinking that maybe instead of adding in an index here, that we should have a little bit of processing on the data before we put it in to the VersionControlObject - to convert from an Object to an array of Objects (in the right order, with the 'name' as a property instead of the key on the Object)

And then do the reverse when we save changes.

Cos - doing it this way with the indexes is probably not having the desired effect of controlling the order in the actual application unless we go and change the app to use the 'index' property for the order (it is currently assuming that they are in the right order in the object)

TomRadford: https://github.com/stock2shop/console_v2/assets/7515754/6e18921e-18dd-4ca0-9c0d-982b2bd1a646

matthampton-stock2shop: I'm realising that I was wrong about this! sorry!

TomRadford:
@matthampton-stock2shop just checking if there's anything else outstanding here or if you're happy to merge?

matthampton-stock2shop:
I'm happy with this

TomRadford (approved):

PR #284, opened 2023-09-20
Justin krs/product import dropdown

justinengle-krs:
Zeplin: 
image
Implementation:
image


src/lib/popup-menu-option.ts
TomRadford, line 1: just checking if this file is meant to be empty

src/components/input/PopupMenu.svelte
```diff
@@ -0,0 +1,23 @@
...
+  let container: HTMLElement;
+
+  function handleClick(option: PopupMenuOption) {
+    option.onClick();
+    open = false;
+  }
+</script>
+
+{#if open}
+  <div class="py-2 bg-white  absolute top-32 z-20 shadow-md rounded w-fit" bind:this={container}>
```
matthampton-stock2shop, line 16: I'm not sure this is going to be a generic popupmenu if it is in a fixed position based on the screen... Or have I misunderstood?

matthampton-stock2shop: Do we maybe need a library for this?

matthampton-stock2shop, line 16: Other issues: you can't close the popup except by clicking on one of the options

src/components/input/PopupMenu.svelte
```diff
@@ -0,0 +1,23 @@
...
+    {#each options as option}
```
matthampton-stock2shop, line 17: If you scroll a little bit down, and the click on the button, it will draw the popup in the same place - looks a bit odd...

(Nitpick)

image

src/components/types.ts
```diff
@@ -11,3 +14,10 @@ export enum OptionInputType {
   NUMBER = 'Number',
   TEXTAREA = 'Textarea',
 }
+
+type SvelteComponentConstructor = new (...args: any[]) => SvelteComponent;
```
TomRadford, line 18: nit: curious if we need this constructor/what its purpose is?

matthampton-stock2shop (approved):

matthampton-stock2shop:
We are going to merge this so that it doesn't cause trouble with the sveltekit branch

PR #286, opened 2023-09-26
Fix tailwind button sizing

TomRadford:
This PR fixes tailwind-merge's ability to correctly handle `text-*` classes that use font-colour and font-size. 
As per doc on https://github.com/dcastil/tailwind-merge/blob/v1.8.0/docs/configuration.md#theme tailwind-merge is unable to automatically derive theme scales (the custom values that we define in tailwind.config.js) for font sizes. We simply needed to manually define the class group (font-size) for all of our custom text-* size values. 

I also cleaned up some files and updated minor tw version to latest.


src/components/form/Btn.svelte
```diff
@@ -77,19 +77,39 @@
   const sizeClass = onlyIcon ? iconSizeClassMap[size] : sizeClassMap[size];
   const variantClass = variantClassMap[variant];
   const iconContainerClass = iconSize ? iconContainerSizeMap[iconSize] : '';
+
+  console.log(
```
matthampton-stock2shop, line 81: Debugging?

matthampton-stock2shop (changes requested):
Looks reasonable.. just a little debugging left behind...

src/components/form/Btn.svelte
TomRadford: Woops! Removing now...

matthampton-stock2shop (approved):

Issue #274, opened 2023-09-07
Product details: Bug fixes & missing feature implementation


TomRadford:
Bugfixes and missing features implementation for Product Details Page

PR #289, opened 2023-09-27
#274: Product details: Bug fixes & missing feature implementation

TomRadford:
Bugfixes and missing features implementation for Product Details Page.


src/components/Product/ProductMaster.svelte
```diff
@@ -2,24 +2,38 @@
...
   export let onSaveChanges: () => Promise<boolean> = null;
   export let handleUpdate: () => void = () => {};
   export let onBeforeSaveChanges: () => Promise<boolean> = null;
   export let saveProduct = true;
+  export let params: ProductParams;
 
-  const product = getContext<ObjectVersionControl<SystemProduct>>('product');
+  const product = new ObjectVersionControl<SystemProduct>(null);
+
+  const productData = createQuery(['product-get', params.id], async () => {
```
DustinS2S, line 27: It appears that the ID is not being populated for the request, I am not yet sure how this pulls the data, this happens when any other route of the product is called (Any other tab of the product) 

If the base route is called `/product/{{product_id}}` it loads without error.
see console error:
```
proxy.js?v=330bcc88:19 TypeError: Cannot read properties of undefined (reading 'id')
    at instance (ProductMaster.svelte:27:1)
    at init (index.mjs:1891:11)
    at new ProductMaster (ProductMaster.svelte:173:1)
    at createComponent (svelte-hooks.js?v=330bcc88:206:20)
    at targetCmp.$replace (svelte-hooks.js?v=330bcc88:269:15)
    at refreshComponent (proxy.js?v=330bcc88:171:15)
    at ProxyAdapterDom.rerender (proxy-adapter-dom.js?v=330bcc88:77:5)
    at proxy.js?v=330bcc88:408:9
    at Array.forEach (<anonymous>)
    at Object.reload (proxy.js?v=330bcc88:406:15)
```

I hardcoded the ID in line 27 and 28 and the data loaded.

DustinS2S (commented):
I am still working on the testing and review, will provide an update soon

DustinS2S:
This is looking awesome, I found the following that may need discussion:

1. When a simple product (One variant with no options) is opened and the `This product has different options` is selected
image
the system displays the variant with the new option field and wont let you save unless the variant has a option value
The bug comes in when you remove any additionally created variants and then untick the `This product has different options`, this saves but the Product still has a product option, which means that when the page is reloaded it is ticked again.

2. When the Tags block is edited it doesnt seem to trigger the `ObjectVersionControl` which then means something else will need to be edited in order for the save button to open.
image
To add to this, Tags do not save regardless, meaning we can never set tags

3. Not sure if this is part of what needed to be done, but the image upload block is not responding.
image

PR #306, opened 2023-10-31
292: Version Controlled Object enhancements 

TomRadford:
Version Controlled Object enhancements with support for deep transient keys and new `options` param:

- add additional transient keys
- add way to sort arrays so they stay stable
- do a direct object comparison rather than stringifying

All of these changes aim to help diff's to be more reliable and consistent between server and client state


package.json
```diff
@@ -71,18 +72,21 @@
   },
   "dependencies": {
     "@tanstack/svelte-query": "^4.33.0",
+    "@types/omit-deep-lodash": "^1.1.2",
     "chart.js": "^4.3.0",
     "clsx": "^2.0.0",
     "cssnano": "^5.1.11",
     "json-stable-stringify": "^1.0.2",
     "just-debounce-it": "^3.1.1",
+    "lodash-omitdeep": "^1.3.5",
```
matthampton-stock2shop, line 81: Guessing we only need one of these?

src/routes/(app)/customer/[id]/+layout.svelte
```diff
@@ -18,8 +18,27 @@
...
+      transientKeys: ['id', 'client_id'],
+      parsers: [
+        {
+          key: 'addresses',
+          parserFn: (v: Address[]) =>
+            v
+              .sort((a, b) => parseInt(b.address_code ?? '0') - parseInt(a.address_code ?? '0'))
+              .map((address) =>
+                Object.fromEntries(
+                  Object.entries(address).map(([k, v]) => [k, v === '' ? null : v])
```
Comment (matthampton-stock2shop) [inline src/routes/(app)/customer/[id]/+layout.svelte:35]: This feels like it is doubling up on the omit stuff

src/routes/(app)/order/[id]/+layout.svelte
```diff
@@ -26,7 +26,10 @@
   }
 
   //Init product VC object as empty object while we fetch data with query
-  const order = new ObjectVersionControl<SystemOrderDataWrapper>({});
+  const order = new ObjectVersionControl<SystemOrderDataWrapper>(
+    {},
+    { transientKeys: ['tax_lines', 'history', 'shipping_lines'] }
```
Comment (matthampton-stock2shop) [inline src/routes/(app)/order/[id]/+layout.svelte:31]: Are these really transient? (I guess history sort of makes sense - but not sure about the others)

src/models/version-controlled.ts
```diff
@@ -47,12 +86,19 @@ export default class ObjectVersionControl<T> {
...
-    return ['created', 'modified', 'hash'];
+    return ['created', 'modified', 'hash', ...this.additionalTransientKeys];
   }
 
   stringify(o): string {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-return
-    return stableStringify(omitBy(o, (v, k) => !v || this.transientKeys.indexOf(k) >= 0 || isEmptyObjectOrArray(v)));
+    this.parsers.map((parser) => {
+      if (o?.[parser.key]) {
+        o[parser.key] = parser.parserFn(o[parser.key]);
```
matthampton-stock2shop, line 96: I think I'd prefer something that doesn't mutate `o`

matthampton-stock2shop: And I think it might be easier if the function(s) are called for each key and value as we traverse the object - rather than once at the root

src/models/version-controlled.ts
```diff
@@ -47,12 +86,19 @@ export default class ObjectVersionControl<T> {
...
+      }
+    });
+    return stableStringify(
```
matthampton-stock2shop, line 99: I'm thinking that maybe using stringify here is actually getting in the way of what we are actually trying to do - which is to compare objects for equality in a slightly complex way

matthampton-stock2shop (commented):
I'm thinking a little pair programming on this might be efficient - if you are ok with that?

src/routes/(app)/customer/[id]/+layout.svelte
Comment (TomRadford) [reply src/routes/(app)/customer/[id]/+layout.svelte:35]: this is specifically to convert empty strings to nulls since the API will always send back empty strings as nulls on this addresses property

src/routes/(app)/order/[id]/+layout.svelte
Comment (TomRadford) [reply src/routes/(app)/order/[id]/+layout.svelte:31]: so we dont ever mutate them on the client (as far as I can tell)  - so diffing them isn't really needed

src/models/version-controlled.ts
TomRadford: got you... this will allow you to be pretty specific about what you mutate - including deep properties - ie o[order.addresses] = parser.parserFn(o[order.addresses);

src/models/version-controlled.ts
TomRadford: interesting - definitely keen to discuss

matthampton-stock2shop (approved):

matthampton-stock2shop (dismissed):

matthampton-stock2shop (approved):
Tom reviewed over my shoulder

Issue #304, opened 2023-10-31
Fix Product Attributes


TomRadford:
Product attributes editor has a couple of nasty issues that we need to resolve once #292 is resolved

PR #307, opened 2023-11-01
#304 Fix product attributes

TomRadford:


src/components/Product/Attribute.svelte
```diff
@@ -5,48 +5,55 @@
...
+  const onDeleteClick = () => {
     deleteAttribute(attribute);
   };
 
-  $: {
-    if (attribute.isToggle()) {
-      attribute.meta.value = attribute.toggleValue ? 'Yes' : 'No';
-    }
-  }
+  $: attributeIndex = $product.workingCopy.meta.findIndex((att) => att.key === attribute.meta?.key);
```
TomRadford, line 21: @matthampton-stock2shop looks like your IDE isnt looking at our prettier config sadly

src/models/version-controlled.ts
```diff
@@ -60,95 +58,91 @@ export default class ObjectVersionControl<T> {
...
     }
-    return left === right
+    return left === right;
   }
 
   areArraysEqual(keyPath: string[], left, right) {
     if (left.length !== right.length) {
-      console.info("Changed key", {keyPath, left, right})
-      return false
+      console.info('Changed key', { keyPath, left, right });
```
TomRadford, line 131: @matthampton-stock2shop seems like WebStorm doesnt seem to be formatting on our prettier config

matthampton-stock2shop: Ah ... Maybe I need to ask it to autoformat - I haven't been in the habit of using that feature

TomRadford:
@matthampton-stock2shop rendering of input if blank value should be resolved now

Ayano2000 (changes requested):
I have tested basic `CRUD` for `productmeta`, adding and removing meta via templates (removing all meta items for a template does deselect the template).

- If a template is removed, any meta that it may have added to a product stays

- [ ]  Default `metatemplate` values do not show when adding a template

-------------

I believe this is not a part of this PR but:
- I cannot edit any of the existing meta_templates, `Successfully added new however some attribute(s) had an error. Error: Undefined property: stdClass::$name`
- Sometimes the modal for creating/editing meta templates gets cached.
- When adding a `metatemplate` of type dropdown, we cannot add any selectable values. 

@TomRadford I'm not sure all of these issues are related to this PR, but I think showing a templates default values when adding it should be working before we merge this

Ayano2000 (approved):
LGTM

PR #314, opened 2023-11-02
#209 Edit customer addreses

TomRadford:


src/components/Customer/Details/Addresses.svelte
```diff
@@ -9,18 +9,46 @@
...
-    $customer.workingCopy.addresses = [...customer.workingCopy.addresses, newAddress];
-    addingAddress = false;
-  }
+  const editAddress = (address: Address) => {
+    editing = true;
+    addressToEdit = address;
+    openModal = true;
+  };
+
+  const saveAction = () => {
```
mannyqi, line 30: There's a weird little bug where the default address is set to false after clicking the "Edit Address" on the side modal even though it is true.

To reproduce bug:
1. Create existing customer address with "Default Address" set to active
2. Edit the customer and then the default address (note the default address toggle is set to true at this point)
3. Click "Edit Address" and notice the default address tag is now missing
image

I tried to debug it myself but couldn't figure out where the default address was getting lost...

mannyqi (changes requested):
I also just noticed that the AddressList storybook is broken:

image

Looks like [type](https://github.com/stock2shop/console_v2/blob/1727d7c30623ea2bcbe0b512ec02bad393232556/src/components/Customer/Details/AddressList.svelte#L22) is undefined somehow.

src/components/Customer/Details/Addresses.svelte
TomRadford: @mannyqi would you mind giving this another try now. 
You should only be able to set one default and the default shouldn't unset itself on edit. 
Made some silly mistakes including not using a deep copy for the addressToEdit state variable - thus resulting in some strange behaviours - hopefully should be working okay now

Only one strange issue regarding the company name. Will document below

TomRadford: @mannyqi  I noticed that when you set the first billing address, it will set that as the `companyName` on the customer - just confirming this is an intentional side effect?

mannyqi:
@TomRadford sorry, would you mind resolve the conflicts in this branch please? I'd like to get this one out the way for you. Sorry about delay

TomRadford:
@mannyqi  resolved conflicts 👍

mannyqi:
@TomRadford All is working great. Just picked up a small style issue:

image

Maybe this can be fixed in another PR so we can get this one merged so long? If yes, then I'll approve this PR.

mannyqi (approved):

Issue #301, opened 2023-10-30
Switch out native `window.alert`'s with Toast/Notification system


TomRadford:
We need to switch out our `window.alert` system with a nice toast system.

Some options:
https://github.com/wobsoriano/svelte-sonner - this card style animation could integrate great with our UI 
https://svelte-french-toast.com/

Potentially we could use this to replace the inline-style notification that are sometimes hard to see:
![image](https://github.com/stock2shop/console_v2/assets/7515754/3ea24028-b191-4139-998e-570c7cb31f9e)

PR #315, opened 2023-11-03
#301 Add svelte-sonner toast library

TomRadford:
- straightforward toast api to start using when we work on data mutations that have user feedback (to get rid of obtrusive `window.alert`) and provide much nicer UX non-obtrusive toasts
- will apply custom styling at a later stage. This is so we can start trickling in `toast.*` wherever needed - adding customisations later on by overriding the `<Toaster>`'s props


src/routes/+layout.svelte
```diff
@@ -33,6 +34,9 @@
 <svelte:head>
   <title>{SITE_TITLE}</title>
 </svelte:head>
+
+<Toaster closeButton position="bottom-center" />
```
matthampton-stock2shop, line 38: Do you need a toaster in the Storybook warpper?

matthampton-stock2shop (approved):

PR #318, opened 2023-11-06
Simple error handler

TomRadford:
This will serve as an error handler for prod users while we wait for proper error boundaries in Svelte 5.
Note that since we're just loading the `/error` page instead of history API's pushstate, prod users will loose the error in the console. 

This offers nicer ux since the user wont be left with a broken UI - with the unfortunate disadvantage of us not being able to see console errors. This should be okay until we need to use something like Sentry or Logrocket.


src/routes/+error.svelte
```diff
@@ -27,6 +36,12 @@
       <h1 class="text-4xl font-semibold">{$page.status}</h1>
       <p>{$page.error?.message}</p>
     </div>
-    <a href="/" class="w-min self-center"><Btn>Back</Btn></a>
+
+    <div class="w-min self-center flex gap-4">
+      <a href="/" class=""><Btn>Back</Btn></a>
```
matthampton-stock2shop, line 41: Shouldn't this say 'Home' or something like that?

src/routes/error/+page.ts
```diff
@@ -0,0 +1,5 @@
+import { error } from '@sveltejs/kit';
+
+export const load = () => {
+  throw error(500, 'A client-side error has occurred');
```
matthampton-stock2shop, line 4: Is this a trick to get it to redner the error page?

TomRadford: Indeed it is - to render the same consistent error page when the client throws.

src/routes/+error.svelte
TomRadford: Agreed.. or do you think a `history.back()` would be better here?

src/routes/+layout.svelte
```diff
@@ -21,6 +21,27 @@
...
+
+  // Global error handler
+  // ToDo: update when Svelte adds some sort of error boundary:
+  // https://github.com/sveltejs/rfcs/pull/46
+  //
+  // Note: we manually redirect to error page
+  // since goto breaks when the Svelte encounters some errors
+  // We pass ignore this in dev so that
+  const isProd = import.meta.env.PROD;
+  const handleError = () => {
```
matthampton-stock2shop, line 33: I think this should actually give you an error object that can be logged right? Should we log it?

src/routes/+layout.svelte
```diff
@@ -36,6 +36,9 @@
       // instead of just seeing a frozen ui
       return (window.location.pathname = '/error');
     }
+    // Dev Mode: warn us that the error would have
+    // thrown the above redirect (to help us catch errors better)
+    console.error('REDIRECT CLIENT TO ERROR PAGE DUE TO:');
```
TomRadford, line 41: @matthampton-stock2shop As discussed, added this little change to help us catch errors that will cause the error page to show

src/routes/+layout.svelte
matthampton-stock2shop: (maybe we should do the logging in prod and dev - but then just do the redirect on prod)

src/routes/+layout.svelte
```diff
@@ -21,6 +21,27 @@
...
   }
+
+  // Global error handler
+  // ToDo: update when Svelte adds some sort of error boundary:
+  // https://github.com/sveltejs/rfcs/pull/46
+  //
+  // Note: we manually redirect to error page
+  // since goto breaks when the Svelte encounters some errors
+  // We pass ignore this in dev so that
+  const isProd = import.meta.env.PROD;
```
matthampton-stock2shop, line 32: Is this the right way to do this? What about this: https://kit.svelte.dev/docs/modules#$app-environment-dev

Can we use that?

src/routes/+layout.svelte
TomRadford: Dev: 

https://github.com/stock2shop/console_v2/assets/7515754/407da230-2e10-42eb-b86b-fc6a1915232c

Prod: 

https://github.com/stock2shop/console_v2/assets/7515754/e0772ede-a797-4192-baae-76d033370e4c

TomRadford: Sadly we redirect to error page outside of SPA route (with window location) and thus console gets cleared

matthampton-stock2shop: You can actually tick a checkbox to "Preserve Log" or something like that - and then it will keep the log when it navigates...

TomRadford: Cool, so confirming.. you're fine with the console clearing by default in prod? Will have to come up with a different strategy if we every use Sentry (since we'll have to wait for there log posts to run before loading up the new error page) - but I think its okay for now

matthampton-stock2shop (approved):

PR #327, opened 2023-11-22
Product Imports: Support for new /import endpoints, field Selector and deep mappings modal

TomRadford:
- [x] Support new product imports endpoint,
- [x] create a new FieldSelector component
- [x] Implement a modal for selecting/adding meta/price/qty:

image

Implementation:

https://github.com/stock2shop/console_v2/assets/7515754/c8db0c64-dc6e-4536-87e5-2619869cee2d

PR #330, opened 2023-11-30
Customer Imports UI

TomRadford:
UI Implementation for Customer Imports.
Builds atop work done in #327


TomRadford:
Implemented. 

Waiting on `the customer export branch which will be merged to the product export branch shortly` to test this further.

TomRadford:
@flumonion just confirming if we're missing a `customer_active` field from the field endpoint? 
image

TomRadford:
Other than deletions (as mentioned above), customer imports is ready for review.

TomRadford:
Moving back to in progress:

Some feedback from Chris:

you can only import customers to a trade store

if you dont have an active trade store we need to show another message.

like we should have a generic "alert" or some other concept that works like javascript alert, just pretier.

and the onclick can check that?

instead of opening the import

TomRadford:
Trade channels only:
image

TomRadford:
No trade channels:
image

PR #338, opened 2023-12-06
Modal and input fixes

TomRadford:
- Stop TextInput from randomly focusing inputs
- Make sure modals are only mounted from their highest parent (ie: in a `#if`)


src/components/Settings/ProductTemplate.svelte
```diff
@@ -221,16 +215,7 @@
   }}
 />
 
-<CustomModal
-  bind:open={openInternal}
-  {isLoading}
-  onClose={() => {
```
TomRadford, line 227: this was a hack that before I knew I needed to set the `global` modifier on the animation

src/components/input/TextInput.svelte
```diff
@@ -51,7 +51,7 @@
     <label class="label min-w-[150px]" for={id}>{title}</label>
   {/if}
   <input
-    {autofocus}
+    autofocus={autofocus || undefined}
```
TomRadford, line 54: `autofocu={false}` seems to still cause autofocussing - hence I'm unsetting if it's not explicitly true

src/components/Customers/AddCustomerModal.svelte
```diff
@@ -170,6 +170,7 @@
         {/if}
         <div class="text-regular text-brand-action font-bold mb-[16px]">DETAILS</div>
         <TextInput
+          autofocus
```
TomRadford, line 173: this will ensure fist input of the modal is focussed

PR #379, opened 2024-01-08
Logs

flumonion:
Logs screen


src/components/Nav/NavLink.svelte
```diff
@@ -2,68 +2,45 @@
...
+          {/each}
+        </div>
+      {/if}
     </div>
-  {/if}
+  </div>
 </a>
 
 <style>
-  :global(.nav-link.active svg) {
```
TomRadford, line 55: good idea to remove this, I just pushed fb06bc009ea56edeb9cecb66e6af6656ff5289c6  which adds the util classes for active

src/routes/(app)/logs/order-errors/+page.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
```
TomRadford: can we remove this route? looks like you're using URLSearchParams to show errors

src/components/Logs/Logs.svelte
```diff
@@ -0,0 +1,128 @@
...
+
+  const redirect = () => {
+    let params: Record<string, string> = {};
+    for (const [key, value] of Object.entries(searchParams)) {
+      if (value !== '') {
+        params[key] = value;
+      }
+    }
+    let query = new URLSearchParams(params);
+    goto(`?${query.toString()}`);
```
TomRadford, line 96: this makes a lot of sense to me, instead of having to use $page.url.searchParams's set & get methods to modify the parameters - which would be another way to do this.

TomRadford (approved):

TomRadford (approved):

Issue #355, opened 2023-12-07
Customer Attributes


TomRadford:
![image](https://github.com/stock2shop/console_v2/assets/7515754/022e67c7-c32f-4b9e-8f2e-de78e5fff431)

TomRadford:
Should be quite quick (same as products)

TomRadford:
@flumonion I noticed that this customer attributes design has a `locked` functionality with an explainer tooltip - are we implementing this? 

![Image](https://github.com/stock2shop/console_v2/assets/7515754/69cc132f-5608-4dea-b7e5-ffed8f6e82f7)

I see that our existing implementation of *product* attributes do *not* have this locked functionality:
![Image](https://github.com/stock2shop/console_v2/assets/7515754/943ba42c-df17-4ec8-ae40-80c70c339a88)
whereas the design does:
![Image](https://github.com/stock2shop/console_v2/assets/7515754/e7ce72d4-3902-469d-bb14-d07a73f9340f)

Issue #319, opened 2023-11-07
Zod for Type Definition and Validation: Research and Implementation 


TomRadford:
Currently, we're not using any time of validator for input and API validation against our schema.

I propose looking into Zod for schema-validation and type inference. 
ie: All our types are converted to Zod schemas and then inferred back as types.

I didn't think this was feasible until I found this awesome util:
https://www.npmjs.com/package/ts-to-zod - which could make this transition potentially feasible in a week or two...
hence I'm making this ticket to investigate further at a later stage.

TomRadford:
This could be NB to discuss in future, but I don't see proper client-side validation happening in initial release. 
@matthampton-stock2shop keen to get your thoughts on validation for the application? 
 	- maybe we can evaluate where its critical and only implement it there?

TomRadford:
something like Superforms [superforms.rocks/concepts/spa](https://superforms.rocks/concepts/spa/) or svelte-use-form ([noahsalvi/svelte-use-form](https://github.com/noahsalvi/svelte-use-form)) to handle validations and give user feedback on relevant inputs.

TomRadford:
We want to start testing this in the:
- Modals: see if we can use the Form
Do prototypes (two methodologies) on the Add Customer modal. 

CSS Option: 
https://dev.to/deyemiobaa/adding-custom-validation-to-a-form-with-tailwindcss-1e7d#:~:text=In%20CSS%2C%20we%20can%20use,to%20force%20a%20certain%20style

TomRadford:
Need to run over with Chris/Matt on a quick call before adding it everywhere

TomRadford:
We had a run over and decided on:

- no inline rendering of feedback because
- we will use onsubmit of forms to do the validation 
- Label in component

PR #387, opened 2024-01-15
Add Customer Attributes, Melt for Modal, PromptModal and async confirm modal

TomRadford:
#355


src/components/Controls/FieldSelector/DynamicFieldSelector.svelte
```diff
@@ -1,8 +1,8 @@
 <script lang="ts">
   import { createGroupFieldsQuery } from '@/queries/fields.query';
   import ComboboxInternal from '@/components/Controls/Combobox/ComboboxInternal.svelte';
-
   import type { FieldEntity } from '@/models/fields/fields';
+  import { isValidContract } from '@/models/pricing-contracts/pricing-contract';
```
flumonion, line 5: Screenshot 2024-01-18 at 17 01 04

flumonion (changes requested):

src/components/Modal/USE_MODAL.md
```diff
@@ -1,3 +1,4 @@
-# Always use the `Modal.svelte` component.
+# Always use the `Modal.svelte` component for building other modals.
 
+- Use the `AddEditModal.svelte` wrapper component for Add/Edit style modals
```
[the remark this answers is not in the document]
TomRadford: AddEditModal was renamed to prompt modal

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,238 @@
+<script lang="ts">
+  import type ObjectVersionControl from '@/models/version-controlled';
+  import { createQuery } from '@tanstack/svelte-query';
+  import type SystemProduct from '@/models/products/product';
+  import handleOutsideClick from '@/utils/actions/handleClickOutside';
+  import Btn from '@/components/form/Btn.svelte';
```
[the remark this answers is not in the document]
TomRadford: This component is effectively a route that supports two different sets of params (ie: the entity) so that we can easily just pass in the correct entity and entity store to make the appropriate changes. 
image

 
 Moving this logic into the +page.svelte for [entity]/attributes, would make the component slightly harder to use but I think I've found a happy middle that makes things a little less complex:

image

Moving save changes handler and templates up the tree (as props) would put a lot of external complexity in the +page.svelte/user of the Attribute component, so I think I'd like to keep those as is - unless we need to use this Attributes component in other places (which is effectively a page component as mentioned above).

Happy medium:
3085036123abf6c1469ad0a76dc4674631f47572

src/components/Controls/FieldSelector/DynamicFieldSelector.svelte
TomRadford: added a nice eslint rules for unused params that we intentionally want to keep (by prefixing `_`) 24d95fa2ecd72a6ae177c03e4f57f5429a9c4773

src/stories/routes/CustomerAttributes.stories.svelte
```diff
@@ -0,0 +1,34 @@
...
+  export const meta = {
+    title: 'routes/Customer/Attributes',
+    component: Attributes,
+    decorators: [() => ({ Component: CustomerDecorator })]
+  };
+</script>
+
+<script lang="ts">
+  import { Story } from '@storybook/addon-svelte-csf';
+  import { page } from '$app/stores';
```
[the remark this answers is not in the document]
TomRadford: resolved these props as part of 3085036

src/components/Modal/Wrappers/PromptModal.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+  import Btn from '@/components/form/Btn.svelte';
+  import Modal from '../Modal.svelte';
+  import type { ModalPosition, ModalSize } from '../modal';
+
+  export let title: string;
+  export let open = false;
+  export let actionClick: (() => void) | ((parm: any) => void);
```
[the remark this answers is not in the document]
TomRadford: 36fd29c235e2f83c7d2647c9313c5acef525163c

src/components/Modal/Modal.svelte
```diff
@@ -42,38 +40,59 @@
     left: 'justify-start',
     center: 'justify-center items-center p-8'
   };
+
+  const {
+    // We're not using trigger/close here since we can't set use:melt={$trigger} on <Btn>,
+    elements: { portalled, overlay, content, title: titleInternal },
+    states: { open: openInternal } // Open conditional is done in highest parent
+  } = createDialog({
+    onOpenChange: ({ curr, next }) => {
```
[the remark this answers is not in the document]
TomRadford: e428ac4b86f3f260ed36b2343fdf275ca35a6716

src/components/Modal/Wrappers/PromptModal.svelte
```diff
@@ -0,0 +1,48 @@
...
+  export let actionText = 'Add';
+  export let position: ModalPosition = 'right';
+  export let size: ModalSize = 'narrow';
+
+  /**
+   * Whether the user can click the add/edit button
+   */
+  export let disabled = false;
```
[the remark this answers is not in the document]
TomRadford: This is for input, made a story: 8d506e577c5919df2e3c95a8a233c60f502a6364  
Adding a confirm modal too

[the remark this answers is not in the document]
TomRadford: Confirm modal:
37a2a6ab9e9b5aabc06c59432cda1d9a987fbbfa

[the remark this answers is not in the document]
flumonion: These are both looking good.
Small note on the styling.

Screenshot 2024-01-22 at 14 36 12

remove the bottom border.
The vertical padding for the inner content should be equal on top and bottom.
My example is wrong...

Screenshot 2024-01-22 at 14 27 52

Since the content text in confirm is likely to be short, perhaps center align text?

src/components/Controls/FieldSelector/DynamicFieldSelector.svelte
flumonion: great.
Out of interest, why would you want to keep an unused param?

i.e.
this
```
export let selectCallback = (_: string) => {};
```
vs
```
export let selectCallback = () => {};
```

src/components/Modal/Wrappers/PromptModal.svelte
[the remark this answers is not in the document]
TomRadford: The second image, PromptModal with center position prop, takes it's own slot so aligning text will happen in there :) 
image
The only case when this is used (ie: with center prop) is for ConfirmModal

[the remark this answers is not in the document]
TomRadford: 432f035229fe95082fcfe8edc229f05a7925badb 

image

src/components/Controls/FieldSelector/DynamicFieldSelector.svelte
TomRadford: I did it as a sort of shorthand, 
instead of 
`export let selectCallback: ((val: string) => void) | undefined;`
and then having to:
```
if (selectCallback) {
	selectCallback(newValue); 
}
```

src/components/Modal/Wrappers/PromptModal.svelte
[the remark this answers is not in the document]
flumonion: nice!

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,232 @@
...
+  import { createCustomerQuery } from '@/queries/customer.query';
+
+  export let entity: FieldEntity;
+  export let meta: Meta[];
+  export let id: number;
+  // Only render template data/ui if entity is product
+  const isProduct = entity === 'product';
+
+  $: productData = createProductQuery(id, isProduct);
+  $: customerData = createCustomerQuery(id, !isProduct);
```
matthampton-stock2shop, line 33: RAther have a prop to tel it to render the skeleton

src/components/Connectors/ChannelMaster.svelte
```diff
@@ -163,11 +164,8 @@
   $: changesMade = $settings.changesMade || settingsScreenChangesMade || $channel.changesMade;
 
   async function saveChanges() {
-    // eslint-disable-next-line no-restricted-globals
-    if (
-      settings.isStale &&
-      !confirm('The settings have changed, by saving you will override these changes.')
-    ) {
+    console.log($settings.isStale);
```
flumonion, line 167: remove log?

src/components/Attributes/Attributes.svelte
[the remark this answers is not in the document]
flumonion: it seems we have lost the templates in the product attributes story?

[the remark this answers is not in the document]
flumonion: The attribute stuff is looking good

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,196 @@
...
+  import type { Writable } from 'svelte/store';
+  import type { FieldEntity } from '@/models/fields/fields';
+  import type TemplateWithItems from '@/models/product-templates/template-with-items';
+
+  export let entity: FieldEntity;
+  export let meta: Meta[];
+  export let isLoading: boolean;
+  export let templateItems: TemplateWithItems[] | undefined = undefined;
+
+  // Only render template data/ui if entity is product
```
flumonion, line 25: small nit...
would you only render template ui if templateItems is not undefined?
just mentioning it...

TomRadford: fixed to only render the `Assign Templates` button when templateItems is assigned 8e828669faddc017fc97b2a834e46ffcbc565943 
Hope that's what you meant.

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,198 @@
+<script lang="ts">
+  import handleOutsideClick from '@/utils/actions/handleClickOutside';
+  import Btn from '@/components/form/Btn.svelte';
+  import IconExpandArrow from '@/components/Icons/IconExpandArrow.svelte';
+  import IconAdd from '@/components/Icons/IconAdd.svelte';
+  import Table from '@/components/grids/ui/Table.svelte';
```
flumonion, line 6: Still got some little gremlins here, sorry.

When deleting the "custom attribute" in your story, two attributes get deleted (string attribute and custom attribute)
Screenshot 2024-01-24 at 08 21 13
Screenshot 2024-01-24 at 08 21 19

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,198 @@
...
+  import type { TemplateDropdownItem } from '@/lib/s2s/types';
+  import AttributeModal from '@/components/Attributes/AttributeModal.svelte';
```
flumonion, line 8: same as previous comment.
When adding a template then deleting one of the items from the template, all items in the template get removed.

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,198 @@
...
+      }
+    }
+  }
+
+  const deleteAttribute = (attribute: Attribute) => {
+    meta = meta.filter((m) => m.id !== attribute.id);
+  };
+
+  const addAttribute = (newAttribute: Meta) => {
+    meta = [...meta, newAttribute];
```
flumonion, line 110: When adding a new attribute, if the keys match we end up with duplicates.
The `meta.key` is unique for any array of metas.

The desired behaviour is that it would update the list.

It might make sense to use a map and not a list for your meta?
for key comparison you could lowercase and trim the keys...

Screenshot 2024-01-24 at 08 27 53
Screenshot 2024-01-24 at 08 28 01

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,198 @@
...
+    toggleModal();
+  };
+
+  const onSaveChanges = getContext<Writable<(() => Promise<boolean>) | undefined>>('onSaveChanges');
+
+  $onSaveChanges = async () => {
+    const requiredWithMissing = attributes.filter(
```
flumonion, line 117: When an attribute is required and you try save it creates some popup below.
What would be nice is if we could introduce whats in the figma design for input errors.

Then the user knows what is missing.
Don't have to do it now but I feel this is a better user experience than the popup (used to show success normally or some fatal error)

Screenshot 2024-01-24 at 08 40 45

src/components/Attributes/Attributes.svelte
```diff
@@ -0,0 +1,198 @@
...
+  import { getContext, onDestroy } from 'svelte';
+  import Skeleton from '../Loader/Skeleton.svelte';
+  import type { Writable } from 'svelte/store';
+  import type { FieldEntity } from '@/models/fields/fields';
+  import type TemplateWithItems from '@/models/product-templates/template-with-items';
+
+  export let entity: FieldEntity;
+  export let meta: Meta[];
+  export let isLoading: boolean;
+  export let templateItems: TemplateWithItems[] | undefined = undefined;
```
flumonion, line 23: another small nit (which you can choose to use or lose)..

The code in this component would be a lot simpler if the templates stuff was another component, `AttributeTemplates`

e.g.
```
{#if templateItems}
<AttributeTemplates {templateItems} bind:meta />
{/if}
```
It renders the two template buttons.

Its job would be to add to the meta dictionary.
unchecking a template would remove all the meta from the dictionary if their value unassigned.

If there is shared functions between the `AttributeTemplates` and `Attributes` components, you could create a `attributes.ts` which has the data types and the functions to add/remove etc...

This composition idea is my preference but i am also happy to leave it as is.

flumonion (changes requested):

src/components/Attributes/Attributes.svelte
TomRadford: Added some checks:
fec518b9a526105210ccf7062deb033d3483926f
29eebe8d8dd5e4a3f0bba5f57a9d32e32d2e4f26

src/components/Attributes/Attributes.svelte
TomRadford: 8006bc70ef39c5acd9af96fe9f7534e79c7c0bb6

src/components/Attributes/Attributes.svelte
TomRadford: 8006bc70ef39c5acd9af96fe9f7534e79c7c0bb6

src/components/Attributes/Attributes.svelte
TomRadford: Agreed, we're not doing validations yet but me and @matthampton-stock2shop discussed potentially using something like Superforms https://superforms.rocks/concepts/spa/ to handle validations and give user feedback on relevant inputs.

TomRadford: 3d09534b38449a8a00dac1c06aabadcd27d10a47 - made a quick change to list the required items

src/components/Attributes/Attributes.svelte
TomRadford: 4c142eef0eef923b2a0fc588a894325eb9659fe9 cleans things up nicely :)

src/components/Attributes/Attributes.svelte
flumonion: > Agreed, we're not doing validations yet (its P3 #319) so when we get there using something like superforms in spa mode https://superforms.rocks/concepts/spa/ will do the trick

perhaps, but this feels like overkill?
Could there just not be a state on our input controls that is an error state?

anyways, good for time being.

flumonion (approved):
yay!

PR #395, opened 2024-01-22
Tooltip button and Icon fixes

flumonion:


src/components/Controls/Button.svelte
```diff
@@ -0,0 +1,119 @@
+<script lang="ts">
```
TomRadford, line 1: Looks good!
So this `Button` will slowly replace the old `Btn`?

src/components/Icons/IconAdd.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  export let size = '20';
+  export let size: string = '';
```
TomRadford, line 2: nit: this type is also inferred when you init it as a string.

TomRadford (approved):

PR #396, opened 2024-01-23
logs to table, button sizing and icons

flumonion:


src/components/Controls/Text.svelte
```diff
@@ -0,0 +1,34 @@
...
+
+<!--
+@component
+render text with useful features, like tooltips
+-->
+<div use:melt={$trigger} class={customClass ?? ''}>{text}</div>
+
+<!-- Tooltip Content-->
+{#if tooltip && $open}
+  <div use:melt={$content} class="z-10 rounded bg-neutral-700 shadow">
```
TomRadford, line 30: Since we're using this tooltip in Button too, we could make a TooltipContent or TooltipInternal:
```svelte
<script lang="ts">
  import { createTooltip, melt } from '@melt-ui/svelte';
  import { fade } from 'svelte/transition';

  // At the expense of initialising an unused tooltip builder,
  // this allows for easy type inference
  const {
    elements: { content: contentInternal, arrow: arrowInternal }
  } = createTooltip();

  export let content = contentInternal;
  export let arrow = arrowInternal;
</script>

<div
  use:melt={$content}
  transition:fade={{ duration: 100 }}
  class="z-10 rounded bg-neutral-700 shadow"
>
  <div use:melt={$arrow} />
  <p class="px-4 py-1 text-regular font-normal leading-6 text-white"><slot /></p>
</div>

```

Then you could use it here as follows:
```svelte
<!-- Tooltip Content-->
{#if tooltip && $open}
  <TooltipContent content={content} arrow={arrow}>{tooltip}</Tmp>
 {/if}
```

src/components/Icons/Icon.svelte
```diff
@@ -1,20 +1,31 @@
...
+    // more meaningful names
+    'i-3': 'h-3 w-3',
+    'i-3.5': 'h-3.5 w-3.5',
+    'i-4': 'h-4 w-4',
+    'i-5': 'h-5 w-5',
+    'i-6': 'h-6 w-6',
+    'i-7': 'h-7 w-7',
+    'i-8': 'h-8 w-8',
+    'i-9': 'h-9 w-9',
+    'i-10': 'h-10 w-10'
```
TomRadford, line 27: makes sense to me 👌

TomRadford (changes requested):

src/components/Controls/Button.svelte
```diff
@@ -94,15 +99,15 @@
   )}
```
TomRadford, line 99: I'm curious how we can set w-full for the button when we want it to fill a container.

TomRadford (changes requested):

src/components/Controls/Text.svelte
flumonion: nice, good find!

https://github.com/stock2shop/console_v2/pull/396/commits/92c82f7c35aec629be1e0139d227c60a75963f04

Lets monitor to see if creating two instances of tooltip is a performance issue.

src/components/Controls/Button.svelte
flumonion: added example and custom class back

https://github.com/stock2shop/console_v2/pull/396/commits/128321409756535c52e3ada074310dbd98654dd4

TomRadford (approved):

PR #428, opened 2024-01-26
Typegen Proposal

TomRadford:
When we have the time post launch, this PR might be cool to investigate further since it add a `generate` scripts that:
1. Updates swagger docs from 1.2 to 2 (for kubb)
2. Fixes reference issues in swagger docs (where we haven't declared a schema for payload/reponse)
3. Generates types for all of our endpoints (using kubb)
4. Generates Tanstack Query mutations/queries w/Axios requests for all of our endpoints
5. 🥳

This would be a bit of work to get working nicely but could be a nice way to ensure our endpoints/typings are as per the spec


src/api-docs/remove-missing-refs.js
TomRadford, line 1: we have to do this since our swagger docs are a bit broken (missing some refs)

PR #436, opened 2024-01-30
Add Support for segment_name and segment_group in DynamicFieldSelector

TomRadford:

PR #449, opened 2024-02-06
Tomr/stub customer segments

TomRadford:


TomRadford:
This PR should be irrelevant one #436 is implemented

PR #455, opened 2024-02-07
Fix Setting Xero Tenants / OAuth2 crash

TomRadford:


TomRadford:
@Ayano2000 this flow seems to work as expected now:
https://github.com/stock2shop/console_v2/assets/7515754/baf38e1a-a6d5-40fa-8a49-1814646cf472

Previously there was a bug with the Tenant selector not picking up the selected tenant correctly

PR #457, opened 2024-02-08
rename SearchFacets title to Search Facets

TomRadford:


mannyqi:
@TomRadford the issue of the "You have unsaved changes. Would you like to discard them?" alert showing even when not making changes is still happening for the "Domain" section. It seems to have been resolved for all the other sections.

PR #458, opened 2024-02-08
Channel Settings: Checkout & Pricing boolean fixes

TomRadford:
Was making the incorrect assumptions of boolean values that were `undefined` here. Should be `true` | `false`


mannyqi (approved):
The "Hide Tax" and "Display Billing Address" both work.

It seems that even though there is a "Tax rate shipping" setting, it is not used anywhere.

PR #462, opened 2024-02-12
Cleanup Login route

TomRadford:


flumonion:
I think we should not worry about design for the time being.
We will revisit this before launch, designer can give feedback.

flumonion (approved):

PR #466, opened 2024-02-15
Set the cookie via a web route instead - so it gets set on the right …

matthampton-stock2shop:
…domain.


TomRadford:
@matthampton-stock2shop happy to close due to #490  ?

PR #469, opened 2024-02-19
Product Details Fixes

TomRadford:


Ayano2000 (approved):
Everything works ( that was mentioned in the issue )

WRT the warehouse stock, I actually do feel kinda weird about the `+, -, .` characters resetting the existing value to `0`. I feel like it might be annoying from a user perspective, especially if they've been updating many fields..

can we get a second opinion before we merge, I'll approve in the meantime though

Ayano2000 (approved):
LGTM

PR #481, opened 2024-02-26
Make price tier & warehouse clearable

TomRadford:


Ayano2000 (commented):
The error when creating a customer is gone, and I don't need to select a warehouse, but I still need to select a default price tier

Ayano2000 (approved):
👍

PR #490, opened 2024-02-29
Auth using `token` search param

TomRadford:


src/routes/(app)/+layout.ts
```diff
@@ -1,15 +1,36 @@
...
+export async function load({ url }) {
+  // We cant use the $page store (and by extension our tokenParam store in load functions)
+  const tokenParam = url.searchParams.get('token');
+  const tokenValue = tokenParam || get(token);
+  if (isString(tokenValue) && tokenValue !== '') {
+    // This will throw 401 if the token is invalid
+    const user = await getValidTokenByToken(tokenValue);
+    /**
+     * @see https://kit.svelte.dev/docs/state-management#no-side-effects-in-load
+     * We'll need to refactor if we ever go SSR
```
TomRadford, line 26: We have a load function here (instead of just doing this all in +layout.svelte) to stop the logged-in application from rendering at all if there's no token/token is invalid. 

Sadly, if the docs don't recommend this IF we ran this code on a server (since we can't set stores in SSR)... 
Since this is unlikely, we should be good to go! 
I just left the message in here in case that ever happened.


## I tried using the load function anyway (as docs recommended) and the results were as follows:

I tried moving this up to root (ie: not in `(app)`) and then returning system_user from load function into +layout.svelte. 

This worked well... however, +layout.svelte was calling the data prop and thus the getValidTokenByToken on every single route change. 

After mount, we'll catch invalid tokens and redirect to login using lib/api/api.ts anyways, so doing this check every time from the load function would be redundant.

src/utils/stores/searchParams.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @todo refactor existing stores binded to URLSearchParams
+ *
+ * Note: Since these depends on SvelteKit's page store,
+ * these utils DO NOT WORK IN LOAD FUNCTIONS.
+ * In the rare occasion you're using this in a load function
+ * please use the `url` param supplied from the load function.
+ *
+ */
+import { queryParam, ssp } from 'sveltekit-search-params';
```
flumonion, line 13: Small nit here. I don't see stores as a util.
Utils for me are things that are independent of project.
I.e. something you could cut and paste into a different project...

Maybe this should live in a `src/stores` dir or in `src/env`?

flumonion: Furthermore, things have got a bit muddled up by other developers as to my original idea behind `lib/s2s`. The idea here was that s2s is a library that could be used in other projects (say b2b store) that is our api and data models. Since we now use Tanstack to do this, this is kinda redundant.

We should probably move all the stuff left in `lib/s2s` out into `queries`, `repos` and `models` dirs, so it all works the same?

Not saying change it here, just saying lets chat about cleaning this up along with all the weird env stuff we have going on...

flumonion (changes requested):

flumonion (approved):

PR #491, opened 2024-03-01
Bind Log & Queue query params to stores

TomRadford:


src/components/ES/Aggregation.svelte
```diff
@@ -5,11 +5,12 @@
...
   import Skeleton from '@/components/Loader/Skeleton.svelte';
   import IconClose from '@/components/Icons/IconClose.svelte';
-  import { createEventDispatcher } from 'svelte';
+  import type { Writable } from 'svelte/store';
 
-  const dispatch = createEventDispatcher();
   export let isLoading: boolean;
   export let agg: Agg;
+  // Typed like this to remain generic
+  export let paramsStore: Writable<Record<string, string | undefined>>;
```
flumonion, line 13: Our ES components now require url state?
Should url state not be managed by a page layout or a component that ties together the different es components?

What if we wanted two queries on one page?

I am happy for you to remove events in favour of a callback, but not sure they should control the url directly.

For the logs, the log component can do all of that?
i.e. if a filter is removed, it can remove from url?

flumonion: What do you think about the `logs.svelte` binding `aggs`?
Means we don't need a callback (or event).
it can then do:-

```ts
$: if(aggs) {
    // change $logParamsStore
  }
```

Basically, the same as you have done with the search.

Other ES components we will have in the future (that are not using url):-

`Results` -> using table
`Pagination` -> using table pagination
`Sorting`

and so on...

src/components/ES/Aggregations.svelte
```diff
@@ -21,14 +22,11 @@
...
   const resetFilters = () => {
-    for (let i = 0; i < aggs.length; i++) {
-      aggs[i].value = '';
-    }
-    filterChanged();
-  };
-
-  const filterChanged = () => {
-    dispatch('filtersChanged', aggs);
+    aggs.forEach((_, i) => {
```
flumonion, line 25: same as above comment. Maybe this should be controlled by the parent component?

src/components/Logs/Logs.svelte
```diff
@@ -59,40 +54,20 @@
...
-      searchParams[agg.key as unknown as LogSearchParam] = agg.value;
-    }
-    redirect();
-  }
-
-  // redirect on search
-  const search = (event: CustomEvent<string>) => {
-    searchParams.search = event.detail;
-    redirect();
+  const onSearch = (search: string) => {
```
flumonion, line 66: yes, i like this. very clean....
add in same for aggs and anything that changes based on url.

If you wanted to make this page easier to test with storybook, you could move all of this one level up to the `routes/logs.svelte`. It feels to me that is where all the routing and loading of data belongs.

But that is just a preference, probably fine where it is for now?
What do you think?

src/models/logs/log-search-params.ts
```diff
@@ -2,12 +2,14 @@ export type LogSearchParam = keyof LogSearchParams;
...
   size: string;
-  search: string;
-  level: string;
-  entity: string;
-  source_id: string;
-  channel_id: string;
-  direction: string;
-  tags: string;
-  origin: string;
+  search?: string;
```
flumonion, line 5: do we still need this type if we have the new query params one?

src/utils/stores/queryParams.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Note: Since these depends on SvelteKit's page store,
+ * these utils DO NOT WORK IN LOAD FUNCTIONS.
+ * In the rare occasion you're using this in a load function
+ * please use the `url` param supplied from the load function.
+ *
+ */
+import { queryParam, queryParameters, ssp } from 'sveltekit-search-params';
+
+export const tokenParam = queryParam('token', ssp.string());
```
flumonion, line 13: yes, i much prefer this file being called what it is :)

flumonion (changes requested):
The code all makes sense and it does make it easier working with url params, great!
Just a few minor things on the ES component composition.

src/utils/stores/queryParams.ts
```diff
@@ -0,0 +1,39 @@
...
+export const urlParam = queryParam('url', ssp.string());
+export const logParams = queryParameters(
+  {
+    from: ssp.string('0'),
+    size: ssp.string('10000'),
```
flumonion, line 18: OOPS!

flumonion (changes requested):

src/utils/stores/queryParams.ts
TomRadford: aa39a69de6e95c441e1dd376458af1ceced4a4ae

src/models/logs/log-search-params.ts
TomRadford: 6c0c5570aaf5487d71ae11b6d5cea4b6a94a2c1f

Just infered rather

src/components/ES/Aggregation.svelte
TomRadford: We no longer make any changes to `aggs`, (in the value property) since we now make these value changes in the store. I can try to add value back and then change $logParamStore value in Logs. But I'm not sure that's what you mean. 

Also fyi as long as URLSearchParam key's are not the same we can have as many queryParameters stores as we like.

TomRadford: Moved the store out of aggs, making things nice and simple using the add/remove/reset callback props 10f230e4148042c8f8de9c5c35cbc9f17c21594d

src/models/es/aggregations.ts
```diff
@@ -1,9 +1,14 @@
...
+/**
+ * Aggregations metadata type.
+ * Use a queryParams store to set/get values reactively
+ *
+ * @see logParams query store for example
+ */
 export interface Agg {
   key: string;
   title: string;
-  value: string;
```
flumonion, line 6: Are you wanting to remove the value from the aggregation, even though it no longer is bound to query params?

It useful to know if an agg has a value set (especially if we want to build ES queries from it) that is why I added value there.

p.s. The story for this no longer works when clicking a agg.

Maybe you want to set the list of query params from the aggs like you do with the search and the other query params?
You could 2 way bind aggs.

flumonion (changes requested):
Looking good. Just my point about the values in the aggs.
Probably worth having a look before merge...

TomRadford:
Closing this off due to https://github.com/stock2shop/console_v2/pull/502

TomRadford:
Closing this off due to https://github.com/stock2shop/console_v2/pull/502

PR #499, opened 2024-03-08
Truncate long column names

martins2s:
This PR adds a change to truncate column names when their width exceeds that of the columns in the import table:

Before:
image

After:
image

I had to update the z-index of the tooltip to a large value, otherwise it wouldn't display on modals. I extended the tailwind config for this mostly to experiment. Let me know if this needs to change.


tailwind.config.js
```diff
@@ -71,6 +71,9 @@ export default {
             transform: 'translateX(100%) scaleX(0.5)'
           }
         }
+      },
+      zIndex: {
+        max: 999999999
```
TomRadford, line 76: this is a great idea!

TomRadford (approved):
Looks good!

PR #504, opened 2024-03-11
Make export modal wording entity specific

martins2s:


TomRadford (changes requested):
nit: maybe `file download`?

TomRadford (approved):

PR #512, opened 2024-03-13
Localise dates

### LINKED ISSUE #361: Dates need to be displayed in their timezone
Tasks:
- Scanning through pages where we render dates and times 
- Checking that we're rendering them in consistent format
- Checking that we're rendering them in the user's timezone consistently 
- Make sure all the dev docker containers are === UTC + 0

Implementation:
- Parsing UTC times into local. 

Concern: 
- Our Ubuntu server is set in UTC (in prod?). 
- Matt tried his darndest to make sure all dev docker container's local time === UTC + 0

Goal:
- We want to parse into the browser's local time 🕰️

martins2s:
Closes #361 

The following table shows all the current date formats used in the console:

display|action|screen|
|---|---|---|
2024-03-12 13:15:46|**Update other long dates to use this format**|
2024-03-11, 08:30:51|Remote ,
5th March 2024|Change to format in first row|imageimage
Tue 5:32PM|**Used in alerts and logs, so keep**
Mon Mar 11 2024 06:30:51|Change to format in first row|imageimage
Tue 12|**Used by x-axis labels in dashboard charts, so keep**

There may be a chance that I've missed something, so please check and let me know, thanks!


src/lib/date-utils.ts
```diff
@@ -76,13 +55,37 @@ export const getLocalShortTime = (isoDateStringWithoutZoneOrTimestamp: string |
...
-export const getLocalDate = (isoDateStringWithoutZoneOrTimestamp: string | number): string => {
+export const getLocalDateTime = (
+  isoDateStringWithoutZoneOrTimestamp: string | number | undefined
+): string | undefined => {
+  if (!isoDateStringWithoutZoneOrTimestamp) {
+    return undefined;
+  }
+
   const date = getDateFromStringOrTimestamp(isoDateStringWithoutZoneOrTimestamp);
-  return date.toISOString().replace('Z', '').replace('T', ' ').replace('.000', '');
```
martins2s, line 85: The problem with .toISOString() is that is creates a date string at UTC, instead of in the user's timezone. I decided to rather use the formatDateTime (formerly formatDate) function at the top of the file to create the date string.

src/lib/date-utils.ts
```diff
@@ -1,41 +1,18 @@
...
-export const dateWithOrdinalDay = (date: Date): string => {
-  const day = date.toLocaleString('en-US', { day: 'numeric' });
-  const ordinalDay = getOrdinalNum(parseInt(day, 10));
-  const monthYear = date.toLocaleString('en-US', {
-    month: 'long',
-    year: 'numeric'
-  });
-  return `${ordinalDay} ${monthYear}`;
-};
+    .replace(/(\d+)\/(\d+)\/(\d+),?/, '$3-$1-$2')
```
martins2s, line 13: Add ,? to regex to remove comma from resulting date string.

src/lib/date-utils.ts
```diff
@@ -76,13 +55,37 @@ export const getLocalShortTime = (isoDateStringWithoutZoneOrTimestamp: string |
   return shortTime.format(date);
 };
 
+/**
+ * Returns a local day time string in the format EEE hh:mm, e.g. Tue 12:34
+ */
+export const getLocalShortDayTime = (
```
martins2s, line 61: I created a function to combine ```getLocalShortDay``` and ```getLocalShortTime``` since in most cases they got used together anyways.

src/lib/date-utils.ts
```diff
@@ -1,41 +1,18 @@
...
       day: '2-digit',
       hour: '2-digit',
       minute: '2-digit',
       second: '2-digit',
+      fractionalSecondDigits: 3,
       hour12: false
     })
-    .replace(/(\d+)\/(\d+)\/(\d+)/, '$3-$1-$2');
-
-const getOrdinalNum = (number: number) => {
```
martins2s, line 14: We're no longer using this cool function, but happy to keep it if any prefers.

src/lib/date-utils.ts
```diff
@@ -76,13 +55,37 @@ export const getLocalShortTime = (isoDateStringWithoutZoneOrTimestamp: string |
...
+  isoDateStringWithoutZoneOrTimestamp: string | number
+): string => {
+  return `${getLocalShortDay(isoDateStringWithoutZoneOrTimestamp)} ${getLocalShortTime(isoDateStringWithoutZoneOrTimestamp)}`;
+};
+
+/**
+ *
+ * Returns a local week day string in the format EEE d, e.g. True 12
+ */
+export const getLocalWeekDay = (isoDateStringWithoutZoneOrTimestamp: string | number): string => {
```
martins2s, line 71: Another potential option was to use 
```
const short = new Intl.DateTimeFormat('en', {
    weekday: 'short',
    day: 'numeric'
  });
```
...but the resulting date string ends up being ```12 Tue``` instead of ```Tue 12```, which is the format currently used for the chart x-axis labels.

flumonion (approved):

